PR #1, opened 2015-02-12
Move statsd setup from main rad project

jongilbraith:
In preparation for some changes that will add reporting to statsd from the models.

Once this is merged, I'll open a second request in the main rad project to remove the ``Stats`` wrapper class from there and bump the rad_core version.

@benlovell


benlovell:
:shipit:

PR #2, opened 2015-02-13
Renames the gem and associated stuff

benlovell:
Like an absolute noddy, I forgot to check rubygems.org when I named
this project `rad_core`, and subsequently found out it was taken (by a
project that has been dead since 2011, but let's not go there)

Followed the convention of prefixing the gem `mas-`.

Lesson learned.


spec/dummy/config/application.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails/all'
 
 Bundler.require(*Rails.groups)
-require "rad_core"
+require "mas/rad_core"
```
tomas-stefano: double quotes here.

spec/dummy/config/routes.rb
```diff
@@ -1,4 +1,3 @@
 Rails.application.routes.draw do
-
-  mount RadCore::Engine => "/rad_core"
+  mount MAS::RadCore::Engine => "/rad_core"
```
tomas-stefano: Double quotes here.

spec/dummy/config/application.rb
benlovell: I wanted to pick these all up in another PR. This one just includes the exact changes to accommodate the change in naming.

Also - thanks, Rails generators for your insistence on mixed quotes.

spec/dummy/config/routes.rb
benlovell: Same here.

spec/dummy/config/application.rb
tomas-stefano: Ok. different PR then :).

tomas-stefano:
:+1:

jongilbraith:
:+1:

PR #5, opened 2015-02-16
Geocoded Firms and Subsidiaries

lshepstone:
This drives the work being done in the `geocoding` branch for `rad`.


app/jobs/geocode_firm_job.rb
```diff
@@ -0,0 +1,19 @@
+class GeocodeFirmJob < ActiveJob::Base
+  def perform(firm)
+    results = Geocoder.search(firm.full_street_address)
+
+    if results.any?
+      Stats.increment('radsignup.geocode.firm.success')
```
benlovell: Maybe we can push this into a helper here and use `stat :success` or `stat :failure` and push the duplication of keys down?

app/jobs/geocode_firm_job.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      firm.latitude = results.first.latitude
```
benlovell: As discussed in person - this is a great candidate for pushing into the domain. `Firm#geocode!` or some such. I suggested a bang method as it mutates your `Firm` instance. But I'm easy on that.

app/jobs/geocode_firm_job.rb
```diff
@@ -0,0 +1,19 @@
...
+      firm.longitude = results.first.longitude
+      firm.save(callbacks: false)
+    else
+      Stats.increment('radsignup.geocode.firm.failed')
+
+      firm.latitude = nil
```
benlovell: Obviously the same applies here as the prior comment.

spec/models/firm_spec.rb
```diff
@@ -185,4 +187,67 @@
...
+    context 'when the address is present' do
+      it 'the firm is scheduled for geocoding' do
+        expect(GeocodeFirmJob).to receive(:perform_later).with(firm)
+        firm.save!
+      end
+    end
+
+    context 'when the address has changed' do
+      let(:firm) do
+        create(:firm).tap do |f|
```
benlovell: This can just be `create(:firm, address_postcode: 'ABCD 123')`

app/jobs/geocode_firm_job.rb
```diff
@@ -0,0 +1,13 @@
+class GeocodeFirmJob < ActiveJob::Base
+  def perform(firm)
+    results = Geocoder.search(firm.full_street_address)
+
+    if results.any?
+      Stats.increment('radsignup.geocode.firm.success')
+      firm.geocode!(results.first.latitude, results.first.longitude)
+    else
+      Stats.increment('radsignup.geocode.firm.failed')
+      firm.geocode!(nil, nil)
```
benlovell: How about `Firm#geocode(lat = nil, long = nil)` so this case golfs down to `firm.geocode!`

app/models/firm.rb
```diff
@@ -86,6 +86,36 @@ class Firm < ActiveRecord::Base
...
+    write_attribute(:latitude, value)
+  end
+
+  def longitude=(value)
+    value = value.to_f.round(6) unless value.nil?
+    write_attribute(:longitude, value)
+  end
+
+  def geocode!(latitude, longitude)
+    begin
```
benlovell: I'd favour just saving and skipping callbacks if that's the intention.

app/models/firm.rb
```diff
@@ -118,6 +148,16 @@ def field_order
 
   private
 
+  def geocode_in_progress?
```
benlovell: Obviously we could drop this if you go with the prior comment's recommendations.

benlovell:
:+1:

alexwllms:
:+1:

PR #6, opened 2015-02-20
Don't geocode partially complete firms.

lshepstone:
Working attempt at fixing the issue that arrises when a partially complete Firm (with a blank full street address) is geocoded.

@benlovell

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -145,7 +145,7 @@ class Firm < ActiveRecord::Base
   private
 
   def geocode_if_needed
-    if full_street_address.present? && full_street_address_changed?
+    if valid? && full_street_address_changed?
       GeocodeFirmJob.perform_later(self)
     end
   end
--- a/spec/models/firm_spec.rb
+++ b/spec/models/firm_spec.rb
@@ -302,6 +302,15 @@ RSpec.describe Firm do
       end
     end
 
+    context 'when the firm is not valid' do
+      before { firm.address_line_one = nil }
+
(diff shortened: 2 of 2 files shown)


spec/models/firm_spec.rb
```diff
@@ -302,6 +312,19 @@
...
 
+    context 'when the address is not present' do
+      before do
+        firm.address_line_one = nil
+        firm.address_line_two = nil
+        firm.address_postcode = nil
+      end
+
+      it 'the firm is not scheduled for geocoding' do
+        expect(GeocodeFirmJob).not_to receive(:perform_later).with(firm)
```
lshepstone: `.with` is probably not needed here actually.

app/models/firm.rb
```diff
@@ -109,7 +110,7 @@ def longitude=(value)
   def geocode!(latitude = nil, longitude = nil)
     self.latitude = latitude
     self.longitude = longitude
-    save!(callbacks: false)
+    save!(callbacks: false, validate: false)
```
benlovell: When you're not validating `save!` is unnecessary. It would never raise because the validations aren't being run.

I would favour us knowing before we get into this branch that we're already valid so we don't have to disable validation. That would make more sense, and be more robust.

My suggestion is that you revert this stuff, and just short-circuit the guard around creating the geocoding job with `valid?` as the first operand.

benlovell:
Ah man, this is much nicer. :+1: :ship: :shipit:

alexwllms:
:+1: STS

PR #7, opened 2015-02-21
Replaces `search` with `coordinates`

benlovell:
Also moved a bunch of stuff around to reduce the logic in the job which is now
just responsible for retrieving the coordinates and handing off to the `Firm`.

@lshepstone WDYT? I went a little crazy. I'm perfectly happy to reduce this
down to just what we needed for `coordinates` if you think it was a step too
far?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    mas-rad_core (0.0.5)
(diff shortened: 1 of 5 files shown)


esbel:
:+1:

benlovell:
Closed as this is being dealt with by @lshepstone's recent changes.

PR #8, opened 2015-02-22
Adds lat/long pair to `Adviser`

benlovell:
I'm just stubbing out these values for now, while @lshepstone works on getting
`Adviser` geocoded in the background. Rather than us both land on these fields
independently I decided to get them in real quick.


lshepstone:
:+1:

PR #9, opened 2015-02-23
5853 Geocode Advisers

lshepstone:
The approach is largely the same as that for firms, except that this uses Geocoder's coordinate search and passes the lat/long pair as a single argument to the adviser's `geocode!` method. Also, the seam for the GeocodeAdviserJob spec is limited to calling the `geocode!` method, and doesn't check that the lat/long attributes were set - that's covered in the adviser spec for `geocode!`.

I've left the lat/long sanitising methods in here, we can chat about a better way forward there.

@benlovell

### DIFF
--- a/app/jobs/geocode_adviser_job.rb
+++ b/app/jobs/geocode_adviser_job.rb
@@ -1,0 +1,15 @@
+require 'geocoder'
+
+class GeocodeAdviserJob < ActiveJob::Base
+  def perform(adviser)
+    point = Geocoder.coordinates(adviser.full_street_address)
+    point ? stat(:success) : stat(:failed)
+    adviser.geocode!(point)
+  end
+
+  private
+
+  def stat(key)
+    Stats.increment("radsignup.geocode_adviser.#{key}")
+  end
+end
--- a/app/jobs/geocode_adviser_job.rb
+++ b/app/jobs/geocode_adviser_job.rb
@@ -1,5 +1,3 @@
-require 'geocoder'
-
 class GeocodeAdviserJob < ActiveJob::Base
   def perform(adviser)
     coordinates = Geocoder.coordinates(adviser.full_street_address)
--- a/app/jobs/geocode_adviser_job.rb
+++ b/app/jobs/geocode_adviser_job.rb
@@ -10,6 +10,6 @@ class GeocodeAdviserJob < ActiveJob::Base
   private
 
   def stat(key)
-    Stats.increment("radsignup.geocode_adviser.#{key}")
+    Stats.increment("radsignup.geocode.adviser.#{key}")
   end
 end
--- a/app/jobs/geocode_adviser_job.rb
+++ b/app/jobs/geocode_adviser_job.rb
@@ -2,9 +2,9 @@ require 'geocoder'
 
 class GeocodeAdviserJob < ActiveJob::Base
   def perform(adviser)
-    point = Geocoder.coordinates(adviser.full_street_address)
-    point ? stat(:success) : stat(:failed)
-    adviser.geocode!(point)
+    coordinates = Geocoder.coordinates(adviser.full_street_address)
+    coordinates ? stat(:success) : stat(:failed)
+    adviser.geocode!(coordinates)
   end
 
   private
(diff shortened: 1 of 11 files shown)


app/jobs/geocode_adviser_job.rb
```diff
@@ -0,0 +1,15 @@
+require 'geocoder'
+
+class GeocodeAdviserJob < ActiveJob::Base
+  def perform(adviser)
+    point = Geocoder.coordinates(adviser.full_street_address)
```
lshepstone: `point` should be `coordinate` here I think.

benlovell: I'd say `coordinates`

app/jobs/geocode_adviser_job.rb
```diff
@@ -0,0 +1,15 @@
...
+    point ? stat(:success) : stat(:failed)
+    adviser.geocode!(point)
+  end
+
+  private
+
+  def stat(key)
+    Stats.increment("radsignup.geocode_adviser.#{key}")
```
benlovell: As discussed, it would be better for us to use dots rather than underscores to denote the segments. This way we can wildcard certain sections when building graphs.

app/models/adviser.rb
```diff
@@ -29,6 +29,27 @@ class Adviser < ActiveRecord::Base
 
   validate :match_reference_number
 
+  after_save :geocode_if_needed
+
+  def full_street_address
+    [postcode, 'United Kingdom'].delete_if(&:blank?).join(', ')
```
benlovell: We can just interpolate postcode with United Kingdom here as we know by this point we'd be valid and postcode would never be blank.

app/models/adviser.rb
```diff
@@ -29,6 +29,27 @@ class Adviser < ActiveRecord::Base
...
+  end
+
+  def latitude=(value)
```
benlovell: These accessors and `geocode!` should be extracted and shared across `Firm` and `Adviser` as this stuff is identical. That's the best we can do for the moment, and then we'll revisit later. I'd quite like to punch together a macro-ish thing for this behaviour.

spec/models/adviser_spec.rb
```diff
@@ -90,4 +90,117 @@
...
+
+      it 'does not cast the value' do
+        expect(adviser.longitude).to be_nil
+      end
+    end
+  end
+
+  describe '#geocode!' do
+    let(:adviser) { create(:adviser) }
+    let(:coordinate) { [Faker::Address.latitude, Faker::Address.longitude] }
```
benlovell: `coordinates`

spec/models/adviser_spec.rb
```diff
@@ -90,4 +90,117 @@
...
+      expect(GeocodeFirmJob).not_to receive(:perform_later)
+      adviser.geocode!(coordinate)
+      adviser.reload
+    end
+
+    it 'the adviser is persisted' do
+      expect(adviser).to be_persisted
+    end
+
+    context 'with a valid coordinate' do
```
benlovell: `with valid coordinates`

spec/models/adviser_spec.rb
```diff
@@ -90,4 +90,117 @@
...
+      it 'the adviser latitude is updated' do
+        expect(adviser.latitude).to eql(coordinate.first.to_f.round(6))
+      end
+
+      it 'the adviser longitude is updated' do
+        expect(adviser.longitude).to eql(coordinate.last.to_f.round(6))
+      end
+    end
+
+    context 'with no coordinate' do
```
benlovell: Basically pluralise `coordinate` in all the things :smile:

app/models/geocode.rb
```diff
@@ -0,0 +1,16 @@
+module Geocode
```
benlovell: I'd name this `Geocodable`so it sounds as though you're extending the including classes with that _behaviour_

app/jobs/geocode_adviser_job.rb
```diff
@@ -0,0 +1,15 @@
+require 'geocoder'
```
benlovell: I'd be surprised if this is necessary here.

spec/jobs/geocode_adviser_job_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'geocoder'
```
benlovell: Same here.

benlovell:
:+1: :ship:

PR #10, opened 2015-02-25
Geocode Firms Like Advisers

lshepstone:
An effort to make the geocode implementation for firms consistent with that of advisers.


esbel:
Awesome! :+1:

spec/spec_helper.rb
```diff
@@ -8,6 +8,7 @@
 require 'pry'
 
 Dir[File.join(File.dirname(__FILE__), 'support', '**', '*.rb')].each { |f| require f }
+Dir[File.join(File.dirname(__FILE__), '**', '*_examples.rb')].each { |f| require f }
```
benlovell: Can we drop this in favour of placing the shared examples in `spec/support/shared_examples/` instead?

benlovell:
Just that one thing, otherwise :tada: :boat: :ship: :shipit:

PR #11, opened 2015-02-27
Translatable Reference Data

lshepstone:
Stab at implementing a mixin for reference data models to support translation by way of introducing a `cy_name` attribute. Currently, this adds translatable support to the other advice method model, in anticipation of it's use for #5725.

### DIFF
--- a/app/models/translatable.rb
+++ b/app/models/translatable.rb
@@ -1,7 +1,5 @@
 module Translatable
-  def en_name
-    self.name
-  end
+  alias_attribute :en_name, :name
 
   def localized_name
     case I18n.locale
--- a/app/models/translatable.rb
+++ b/app/models/translatable.rb
@@ -1,0 +1,14 @@
+module Translatable
+  def en_name
+    self.name
+  end
+
+  def localized_name
+    case I18n.locale
+      when :en
+        en_name
+      when :cy
+        cy_name
+    end || name
+  end
+end
--- a/app/models/translatable.rb
+++ b/app/models/translatable.rb
@@ -3,10 +3,10 @@ module Translatable
 
   def localized_name
     case I18n.locale
-      when :en
-        en_name
-      when :cy
-        cy_name
-    end || name
+    when :en
+      en_name
+    when :cy
+      cy_name
+    end
   end
 end
--- a/spec/dummy/config/locales/cy.yml
+++ b/spec/dummy/config/locales/cy.yml
@@ -1,0 +1,2 @@
+cy:
+  hello: "Hello world"
--- a/spec/factories/other_advice_method.rb
+++ b/spec/factories/other_advice_method.rb
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :other_advice_method do
     name { Faker::Lorem.sentence }
+    sequence(:cy_name) { |n| "Dull gyngor arall #{n}" }
   end
 end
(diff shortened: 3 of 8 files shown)


app/models/translatable.rb
```diff
@@ -0,0 +1,14 @@
+module Translatable
+  def en_name
```
benlovell: Can we replace this with `alias :en_name :name`

app/models/translatable.rb
```diff
@@ -0,0 +1,14 @@
...
+    self.name
+  end
+
+  def localized_name
```
benlovell: Indent the `when` at the same level as `case` and instead of `||`ing, have `name` as the `else` within the `case`. E.g.

```ruby
case I18n.locale
when :en
  en_name
when :cy
  cy_name
else
  name
end
```

Also, it may be more defensive to explicitly bomb out if you hit the `else` clause. These kind of silent, soft-failures can cause surprises.

spec/dummy/config/locales/cy.yml
```diff
@@ -0,0 +1,2 @@
+cy:
```
benlovell: Is this file necessary?

spec/factories/other_advice_method.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :other_advice_method do
     name { Faker::Lorem.sentence }
+    sequence(:cy_name) { |n| "Dull gyngor arall #{n}" }
```
benlovell: What's the rationale for making this a sequence?

spec/dummy/config/locales/cy.yml
lshepstone: Yeah, otherwise when specs run it complains about there being no "cy" locale. Any ideas around that?

benlovell: There is, but I guess this is the easiest approach in this case. Ignore that comment :smile:

spec/factories/other_advice_method.rb
lshepstone: Realistically reference data entries should be unique, so this ensures they are - same as you did for the accreditation and professional body models etc.

benlovell: I added the sequence so I could uniquely identify the options by name in the specs. Also, the `name` here isn't a sequence so it is somewhat contradictory.

app/models/translatable.rb
lshepstone: So the intention here was not quite to default to name for an unrecognised locale, but rather to ensure that a value is returned if none was found. I agree though, that behaviour probably is unexpected, so can be removed.

spec/factories/other_advice_method.rb
lshepstone: Just not sure how to generate unique Welsh sentences :smile:

benlovell: :smile: My point is, if we don't care about `name` being unique, we shouldn't be caring about `cy_name`'s uniqueness or lack thereof. The reason I'm being pushy with this is that this kind of thing tends to get cargo-culted around the rest of the project, and we'll end up with sequences where they're not being used for any specific reason.

lshepstone: Sure, no problem - was looking at Faker's support for Welsh. None. I'll revert to the same as `name`. :smile:

benlovell:
:ship: :shipit:

PR #12, opened 2015-02-28
Indexing firms and advisers

benlovell:
Provides a basis for indexing `Firm`s and their associated `Adviser`s.

Some caveats and areas for attention later:

* Robustness around ES communication
* Robustness around indexing
* There may be some timing issues around `Adviser` being geocoded before being indexed

These points will be addressed, but this is good to get some eyes on for now.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    mas-rad_core (0.0.7)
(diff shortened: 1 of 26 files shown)


lshepstone:
Looks solid to me! Nice that you didn't have to use an ES library in the end. I'm also concerned around the timing of the geocoding and indexing, but I think we'll have to address that as it comes. Lekker!

:+1:

PR #16, opened 2015-03-03
Expose other advice methods in FirmSerializer

esbel:
* expose names of other advice methods in FirmSearch to allow elastic search index on them

part of #5725

### DIFF
--- a/app/serializers/firm_serializer.rb
+++ b/app/serializers/firm_serializer.rb
@@ -7,7 +7,8 @@ class FirmSerializer < ActiveModel::Serializer
     :options_when_paying_for_care,
     :equity_release,
     :inheritance_tax_planning,
-    :wills_and_probate
+    :wills_and_probate,
+    :other_advice_methods
 
   has_many :advisers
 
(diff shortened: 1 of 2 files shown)


lshepstone:
Perhaps it'd be better to see this as part of the bigger picture for #5725?

esbel:
Agree. However, that's probably the only change to rad-core required to complete #5725 (an optimist here!).

lshepstone:
Aah, right! So, we need to have a chat about if we're indexing names or IDs for reference data.

esbel:
@lshepstone @benlovell `name`s changed to `id`s, as discussed

lshepstone:
LGTM :+1: Might need a rebase now, though! ;)

esbel:
yes, it did - done.

app/serializers/firm_serializer.rb
```diff
@@ -34,4 +35,8 @@ def wills_and_probate
   def _id
     object.id
   end
+
+  def other_advice_methods
+    object.other_advice_methods.collect(&:id)
```
benlovell: `object.other_advice_method_ids` is more efficient.

esbel: changed.

benlovell:
:ship:

esbel:
cheers @lshepstone @benlovell

PR #17, opened 2015-03-03
Raketask firms:index to provide elasticsearch indexing

esbel:
* rake task to reindex firms
* remove empty .rake file

### DIFF
--- a/lib/mas/tasks/firms.rake
+++ b/lib/mas/tasks/firms.rake
@@ -1,0 +1,14 @@
+namespace :firms do
+  desc 'Index all existing firms'
+  task :index => :environment do
+    puts 'Do you want to index all existing firms? [type `yes` to confirm]'
+    confirmation = STDIN.gets.chomp
+    if confirmation.downcase == 'yes'
+      puts 'Building firms index...'
+      Firm.all.each {|f| IndexFirmJob.perform_later(f) }
+      puts '...indexing done.'
+    else
+      puts 'Indexing aborted.'
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/mas/tasks/firms.rake
```diff
@@ -0,0 +1,14 @@
+namespace :firms do
+  desc 'Index all existing firms'
+  task :index => :environment do
+    puts 'Do you want to index all existing firms? [type `yes` to confirm]'
+    confirmation = STDIN.gets.chomp
```
dougdroper: Out of interest why bother with asking for confirmation? Surly by running this you intend to index?

esbel: it might take a while to index them... that was @benlovell's suggestion

benlovell:
@dougdroper SAFETY FIRST!

safety first

benlovell:
:+1: :ship:

lib/mas/tasks/firms.rake
```diff
@@ -0,0 +1,14 @@
...
+    if confirmation.downcase == 'yes'
+      puts 'Building firms index...'
+      Firm.all.each {|f| IndexFirmJob.perform_later(f) }
```
benlovell: You could use `find_each` to batch these rather than loading them all in memory but we know there won't be too many of these so no biggy.

esbel: totally agree, but I've already merged it after seeing a dog, ship, bike and helmet

will come back to it if we encounter any problems :)

PR #18, opened 2015-03-03
Resolves issue with calculation of `from`

benlovell:
Was incorrectly adding an offset, prior to this change.

The `from` is zero-indexed. So the first result of the second page would be
`10`.


esbel:
:+1:

alexwllms:
LGTM :+1:

benlovell:
:bug: :bomb:

PR #19, opened 2015-03-04
Schedule indexing after geocoding successfully

benlovell:
Also limit serialized advisers to geocoded ones only.

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -12,8 +12,6 @@ class Adviser < ActiveRecord::Base
 
   before_validation :upcase_postcode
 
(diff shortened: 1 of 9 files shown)


app/models/adviser.rb
```diff
@@ -12,8 +12,6 @@ class Adviser < ActiveRecord::Base
 
   before_validation :upcase_postcode
 
-  after_save :schedule_indexing, if: -> { valid? && geocoded? }
```
esbel: :heart_eyes_cat: every time an AR filter is deleted, a new pony is born. :heart_eyes_cat:

esbel:
:+1:

asmega:
:+1:

PR #20, opened 2015-03-05
5838 Pension Pot Size Filter

lshepstone:
Support for indexing investment size IDs to support filtering for a pension pot size.

### DIFF
--- a/spec/serializers/firm_serializer_spec.rb
+++ b/spec/serializers/firm_serializer_spec.rb
@@ -32,9 +32,9 @@ RSpec.describe FirmSerializer do
       expect(subject[:wills_and_probate]).to be
     end
 
-    it 'exposes names of `other_advice_methods`' do
-      expect(subject[:other_advice_methods]).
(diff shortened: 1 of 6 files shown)


spec/serializers/firm_serializer_spec.rb
```diff
@@ -32,9 +32,12 @@
       expect(subject[:wills_and_probate]).to be
     end
 
-    it 'exposes names of `other_advice_methods`' do
-      expect(subject[:other_advice_methods]).
-        to eql(firm.other_advice_method_ids)
+    it 'exposes `other_advice_method_ids`' do
```
esbel: thanks for that :)

esbel:
:+1:

benlovell:
:+1:

spec/serializers/firm_serializer_spec.rb
lshepstone: np! ;)

PR #21, opened 2015-03-09
Serialize the percentage values

benlovell:
We'll need to aggregate over these values, so rather than serializing
just their existence -- actually serialize their values.


esbel:
:+1:

PR #22, opened 2015-03-10
Take ordering into account

benlovell:
When multiple sort filters are applied, the order of the returned sort
values becomes positional. In which case we should always return the
last entry as we will always sort by geographical distance last.

This is somewhat fragile at the moment, so we'll revisit if we apply
more sort filters.

### DIFF
--- a/spec/lib/mas/firm_result_spec.rb
+++ b/spec/lib/mas/firm_result_spec.rb
@@ -33,12 +33,20 @@ RSpec.describe FirmResult do
       expect(subject.name).to eq('Financial Advice 1 Ltd.')
     end
 
+    it 'maps the `total_advisers`' do
+      expect(subject.total_advisers).to eq(1)
+    end
+
     it 'maps the `closest_adviser`' do
       expect(subject.closest_adviser).to eq(0.7794549719530739)
     end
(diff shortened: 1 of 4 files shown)


benlovell:
Will bump and release when it's thumbed.

lshepstone:
:+1:

spec/lib/mas/firm_result_spec.rb
```diff
@@ -33,12 +33,20 @@
...
       expect(subject.closest_adviser).to eq(0.7794549719530739)
     end
 
-    it 'maps the `total_advisers`' do
-      expect(subject.total_advisers).to eq(1)
+    context 'when sorted by types of advice first' do
+      before { data['sort'].unshift(123) }
+
+      it 'maps the `closest_adviser`' do
+        expect(subject.closest_adviser).to eq(0.7794549719530739)
```
esbel: for float comparison, we should really use `be_within(delta).of(value)` matcher

esbel: however, it's the goal of this PR to deal with numerics

esbel:
:+1:

PR #23, opened 2015-03-10
Resolves issue with `total_pages` calculation

benlovell:


lshepstone:
:+1:

lib/mas/paginateable.rb
```diff
@@ -1,6 +1,12 @@
 module Paginateable
   def total_pages
-    (total_records / page_size) + 1
+    if total_records < page_size
+      1
+    elsif (total_records % page_size).zero?
+      total_records / page_size
+    else
+      (total_records / page_size) + 1
+    end
```
esbel: elseif and else condition can be squashed into:

```ruby
else
  (total_records - 1) / page_size + 1
end
```

```
2.2.0 :045 >   [0, 1, 10, 19, 20, 21, 39, 40, 41].each do |p|
2.2.0 :046 >       r = (p-1) / ps + 1
2.2.0 :047?>     puts "#{p} => #{r}"
2.2.0 :048?>   end
0 => 0
1 => 1
10 => 1
19 => 1
20 => 1
21 => 2
39 => 2
40 => 2
41 => 3
```

benlovell: Most of those are incorrect.

esbel: for 10 as page size:

```
0 => 0
1 => 1
10 => 1
19 => 2
20 => 2
21 => 3
39 => 4
40 => 4
41 => 5
```

esbel:
:+1:

PR #25, opened 2015-03-12
5838 Additional Investment Size Filters

lshepstone:
@benlovell @esbel

### DIFF
--- a/app/serializers/firm_serializer.rb
+++ b/app/serializers/firm_serializer.rb
@@ -9,6 +9,7 @@ class FirmSerializer < ActiveModel::Serializer
     :inheritance_tax_planning,
     :wills_and_probate,
     :other_advice_methods,
+    :advises_on_investments,
     :investment_sizes
 
   has_many :advisers
@@ -43,6 +44,10 @@ class FirmSerializer < ActiveModel::Serializer
 
   def other_advice_methods
     object.other_advice_method_ids
+  end
+
+  def advises_on_investments
+    object.investment_size_ids.any?
   end
 
   def investment_sizes
--- a/app/serializers/firm_serializer.rb
+++ b/app/serializers/firm_serializer.rb
@@ -10,7 +10,8 @@ class FirmSerializer < ActiveModel::Serializer
     :wills_and_probate,
     :other_advice_methods,
     :advises_on_investments,
-    :investment_sizes
+    :investment_sizes,
+    :investment_transfers
 
   has_many :advisers
 
@@ -52,5 +53,9 @@ class FirmSerializer < ActiveModel::Serializer
 
   def investment_sizes
     object.investment_size_ids
+  end
+
+  def investment_transfers
+    object.pension_transfer_percent > 0
   end
 end
--- a/spec/serializers/firm_serializer_spec.rb
+++ b/spec/serializers/firm_serializer_spec.rb
@@ -40,6 +40,10 @@ RSpec.describe FirmSerializer do
       expect(subject[:investment_sizes]).to eql(firm.investment_size_ids)
     end
 
+    it 'exposes `advises_on_investments`' do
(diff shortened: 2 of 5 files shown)


app/serializers/firm_serializer.rb
```diff
@@ -45,7 +47,15 @@ def other_advice_methods
     object.other_advice_method_ids
   end
 
+  def advises_on_investments
```
benlovell: No spec for this?

app/serializers/firm_serializer.rb
```diff
@@ -45,7 +47,15 @@ def other_advice_methods
...
+    object.investment_size_ids.any?
```
tomas-stefano: This methods returns true or false. Can we change to use '?' on the end?

app/serializers/firm_serializer.rb
```diff
@@ -45,7 +47,15 @@ def other_advice_methods
...
+  end
+
   def investment_sizes
     object.investment_size_ids
   end
+
+  def investment_transfers
+    object.pension_transfer_percent > 0
```
tomas-stefano: Same here.

app/serializers/firm_serializer.rb
benlovell: It's in a serializer.

app/serializers/firm_serializer.rb
tomas-stefano: Ow, it's a serializer. My bad. But I would move this to the object.

spec/serializers/firm_serializer_spec.rb
```diff
@@ -40,6 +40,10 @@
       expect(subject[:investment_sizes]).to eql(firm.investment_size_ids)
     end
 
+    it 'exposes `investment_transfers`' do
+      expect(subject[:investment_transfers]).to be
```
tomas-stefano: Is there other context for this specs? If so, can we break into contexts?

app/serializers/firm_serializer.rb
tomas-stefano: Yes, my bad :)

app/serializers/firm_serializer.rb
benlovell: `object.pension_transfer_percent.nonzero?`

app/serializers/firm_serializer.rb
lshepstone: Got rebased out like a knob - sorry about that.

benlovell:
:+1:

esbel:
:+1:

spec/serializers/firm_serializer_spec.rb
```diff
@@ -40,6 +40,14 @@
...
     end
 
+    it 'exposes `advises_on_investments`' do
+      expect(subject[:advises_on_investments]).to be
+    end
+
+    it 'exposes `investment_transfers`' do
+      expect(subject[:investment_transfers]).to be
+    end
+
```
tomas-stefano: Will not have specs for when is false?

lshepstone: Short answer is that we build our queries to only ever match against `true` - but yeah, wouldn't hurt to have the additional coverage here.

benlovell: It's already covered. What would checking the inverse give us @tomas-stefano

benlovell: Unless you're talking about both cases, I would be more likely to agree about the first case. But even still, we'd just be checking that rails works correctly, IMO.

tomas-stefano: Is it already covered both cases? I tried to say e.g: if pension_transfer_percent is zero the API will returns false for that field. Or if investment_size_ids is empty will returns false. That both scenarios are covered?

PR #26, opened 2015-03-12
Ensure `investment_transfers` is false when pension transfer percent is null

lshepstone:
@benlovell @esbel @tomas-stefano


esbel:
:+1:

tomas-stefano:
:+1:

PR #34, opened 2015-03-30
6046 Index Firm when it's updated

lshepstone:
Also attempts to make specs for `GeocodeFirmJob` and `GeocodeAdviserJob` consistent and move geocode related methods to the `Geocodable` module.

@benlovell


benlovell:
:+1: nice one. :ship:

PR #39, opened 2015-04-07
Restrict `Principal#firm` to parent only

benlovell:
Ensures this association is scoped to parent firms, and not subsidiaries.


alexwllms:
:+1:

PR #42, opened 2015-04-09
Eliminate ActiveJob race

benlovell:
The jobs were being scheduled on save, and often the referred to instance would
not yet be committed to the database, thus causing a subtle race on retrieval.

By scheduling the jobs on `after_commit` we can eliminate the issue.


alexwllms:
:bow:

amansinghb:
:+1: no tone for my skin

PR #48, opened 2015-04-15
Reworks job enqueing specs

benlovell:
* Removes unnecessary spec around postcode changes. This behaviour
 is now triggered regardless.
* Doesn't isolate against the actual job class.


amansinghb:
:+1:

alexwllms:
:ship:

PR #53, opened 2015-04-28
Fix intermittent test failure in adviser_spec

munckymagik:
`after_commit` hooks are not triggered while testing. FactoryGirl
callbacks were being used to trigger the `geocode` method from
`after_save` instead, but these are only triggered when using
FactoryGirl's `create` call and NOT `build`.

The tests in the 'after commit' block were using FactoryGirl's `build`
method and so when run in isolation (or before other tests in the file)
they did not trigger the callbacks in the factory. Here's a fixed run of
the tests that consistently recreates the failure.

  rspec spec/models/adviser_spec.rb:124 --order random:26438 -f d

The fix we've implemented takes out the callbacks in the factory and
explicitly triggers the `after_commit` callback instead.

### DIFF
--- a/spec/factories/adviser.rb
+++ b/spec/factories/adviser.rb
@@ -15,9 +15,6 @@ FactoryGirl.define do
     confirmed_disclaimer true
     firm
 
-    before(:create) { |a| a.class.skip_callback(:save, :after, :geocode) }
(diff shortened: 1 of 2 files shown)


benlovell:
Nice one 👍

neoeno:
:+1:

PR #54, opened 2015-05-01
Handle destruction of an adviser

munckymagik:
* Prevents an adviser geocoding job being scheduled
* Re-indexes the containing firm
* Change forces Firm#geocode to become public

@neoeno @benlovell

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -31,7 +31,7 @@ class Adviser < ActiveRecord::Base
 
   validate :match_reference_number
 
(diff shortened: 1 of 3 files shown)


benlovell:
:+1: LGTM. Version bump?

munckymagik:
Caden's going to be doing the Firm destruction today I believe so shall we leave the version bump till after then?

benlovell:
:+1:

neoeno:
Looks good :+1:

PR #55, opened 2015-05-01
Add rake task to audit Elastic against the database.

neoeno:
What with all the manual moving advisers and deleting firms we're being asked to do, it would be useful to feel secure that Elastic isn't getting out of sync.

Here's a task that will compare the local database with Elastic, and then go through all the docs in Elastic and check they exist locally.

It could be more efficient, but it runs in OK time.

@munckymagik


munckymagik:
:+1:

lib/mas/tasks/firms/audit.rake
```diff
@@ -0,0 +1,75 @@
...
+      end
+    end
+  end
+
+  def all_firms
+    @all_firms ||= FirmRepository.new.all['hits']['hits']
+  end
+
+  def serialize_firm(firm)
+    JSON.parse(FirmSerializer.new(firm).as_json.to_json)
```
neoeno: Just `.to_json`?

lib/mas/rad_core/engine.rb
```diff
@@ -22,6 +22,11 @@ class Engine < ::Rails::Engine
           FactoryGirl.definition_file_paths << root.join('spec', 'factories')
         end
       end
+
+      rake_tasks do
+        load "tasks/firms/audit.rake"
+        load "tasks/firms/index.rake"
```
neoeno: Quotes!

mconnell:
Other than the comment about whether you need both `.as_json` & `.to_json`, looks good :+1:

PR #58, opened 2015-05-07
Move advisers

munckymagik:
Implement moving single or batches of advisers between firms.

Ensures all records get re-indexed in ES.

Example client code:

```ruby
# Find the advisers you want to move, in this case restricting just to advisers in the one firm
from_firm = Firm.find_by(fca_number: ....)
ref_nums = ['GBP00003', 'RMP01116']
advisers_to_move = from_firm.advisers.where(reference_number: ref_nums)

# Validation for this scenario, e.g.
fail 'not all advisers are in the same firm' if ref_nums.size != advisers_to_move.count

# Move the advisers. `move_all_to_firm` operates only on the `advisers_to_move` result set
receiving_firm = Firm.find_by(fca_number: ....)
advisers_to_move.move_all_to_firm(receiving_firm)
```

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -37,6 +37,14 @@ class Adviser < ActiveRecord::Base
   after_commit :geocode
   after_commit :reindex_old_firm
 
+  def self.move_all_to_firm(receiving_firm)
+    transaction do
+      current_scope.each do |adviser|
+        adviser.update!(firm: receiving_firm)
+      end
+    end
+  end
+
   def full_street_address
     "#{postcode}, United Kingdom"
   end
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -1,6 +1,8 @@
 class Adviser < ActiveRecord::Base
   include Geocodable
 
+  attr_reader :old_firm_id
+
   belongs_to :firm
 
   has_and_belongs_to_many :qualifications
@@ -31,7 +33,9 @@ class Adviser < ActiveRecord::Base
 
   validate :match_reference_number
 
+  after_save :flag_changes_for_after_commit
   after_commit :geocode
+  after_commit :reindex_old_firm
 
   def full_street_address
     "#{postcode}, United Kingdom"
@@ -48,12 +52,23 @@ class Adviser < ActiveRecord::Base
 
   private
 
+  # All record of what changed is gone by the time we get to the after_commit
+  # hooks, so we need to store any important changes here to be actioned later.
+  def flag_changes_for_after_commit
+    @old_firm_id = firm_id_change.first if firm_id_changed?
+  end
+
   def geocode
     if destroyed?
       firm.geocode
     elsif valid?
       GeocodeAdviserJob.perform_later(self)
     end
+  end
+
+  def reindex_old_firm
+    IndexFirmJob.perform_later(@old_firm_id) if @old_firm_id.present?
+    @old_firm_id = nil
   end
 
   def upcase_postcode
--- a/spec/models/adviser_spec.rb
+++ b/spec/models/adviser_spec.rb
@@ -121,7 +121,7 @@ RSpec.describe Adviser do
     let(:job_class) { GeocodeAdviserJob }
   end
 
-  describe 'after commit' do
+  describe 'after_commit :geocode' do
     let(:adviser) { create(:adviser) }
 
     before do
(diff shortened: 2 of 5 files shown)


app/models/adviser.rb
```diff
@@ -32,6 +32,15 @@ class Adviser < ActiveRecord::Base
   validate :match_reference_number
 
   after_commit :geocode
+  after_save :check_for_changes
+
+  def self.move_to(receiving_firm)
+    self.transaction do
```
neoeno: Can we lose the `self.`?

app/models/adviser.rb
```diff
@@ -32,6 +32,15 @@ class Adviser < ActiveRecord::Base
   validate :match_reference_number
 
   after_commit :geocode
+  after_save :check_for_changes
+
+  def self.move_to(receiving_firm)
```
neoeno: Can we think of a better name for this? Maybe something both more descriptive of the operation and something that signals its potential destructiveness. Maybe `move_all_to_firm`? I'm thinking along the lines of `destroy_all` and `update_all`.

app/models/adviser.rb
```diff
@@ -48,12 +57,21 @@ def field_order
...
     if destroyed?
       firm.geocode
     elsif valid?
       GeocodeAdviserJob.perform_later(self)
     end
+
+    if @old_firm_id.present?
+      IndexFirmJob.perform_later(@old_firm_id)
+      @old_firm_id = nil
+    end
```
neoeno: Could this cause a race condition of IndexFirmJob against GeocodeAdviserJob (which calls IndexFirmJob)?

app/models/adviser.rb
```diff
@@ -48,12 +57,21 @@ def field_order
 
   private
 
+  def check_for_changes
+    @old_firm_id = firm_id_change.first if firm_id_changed?
+  end
```
neoeno: Is there a reason we're setting up an instance variable in an `after_save` callback rather than just calling this directly in the `geocode` callback?

app/models/adviser.rb
neoeno: Further, could we put this in a different method with a different callback? Since it's not directly related to geocoding. Or possibly rename this callback to call through to geocode and a method containing this new code.

spec/models/adviser_spec.rb
```diff
@@ -163,4 +163,75 @@ def queue_contains_a_job_for(job_class)
...
+
+  describe 'after_save :check_for_changes' do
+    let(:original_firm) { create(:firm) }
+    let(:receiving_firm) { create(:firm) }
+    subject { create(:adviser, firm: original_firm) }
+
+    def save_with_commit_callback(model)
+      model.save!
+      model.run_callbacks(:commit)
+    end
```
neoeno: Nice.

spec/models/adviser_spec.rb
```diff
@@ -163,4 +163,75 @@ def queue_contains_a_job_for(job_class)
...
+        expect(GeocodeAdviserJob).to receive(:perform_later).with(subject)
+        subject.firm = receiving_firm
+        save_with_commit_callback(subject)
+      end
+
+      it 'triggers reindexing of the original firm' do
+        expect(IndexFirmJob).to receive(:perform_later).once().with(original_firm.id)
+        subject.firm = receiving_firm
+        save_with_commit_callback(subject)
+        subject.run_callbacks(:commit)
```
neoeno: Do we need to run the commit callback twice here?

spec/models/adviser_spec.rb
```diff
@@ -163,4 +163,75 @@ def queue_contains_a_job_for(job_class)
...
+      expect(receiving_firm.advisers.count).to be(5)
+      expect(receiving_firm.adviser_ids).to include(advisers_to_move[0].id,
+                                                    advisers_to_move[1].id)
+      expect(original_firm.adviser_ids).not_to include(advisers_to_move[0].id,
+                                                       advisers_to_move[1].id)
+    end
+
+    context 'when one of the move operations fails' do
+      let(:advisers_to_move) { original_firm.advisers.limit(3) }
+      let(:invalid_record_index) { 1 }
```
neoeno: Would `let(:invalid_record) { advisers_to_move[1] }` be more convenient?

app/models/adviser.rb
munckymagik: No, this is what makes it work on a result set rather than just a single record. For a single record you just do `adviser.firm = new_firm`.

app/models/adviser.rb
munckymagik: No race condition, the latter code is to ensure the OLD firm is updated, the GeocodeAdviser job will only reindex the new firm.

neoeno: You're right — missed that. And the existing geocode method will reindex the current firm come-what-may, as I understand it?

app/models/adviser.rb
munckymagik: Yes as discussed, all record of what changed is gone by the time we get to the after_commit hook.

spec/models/adviser_spec.rb
munckymagik: Ok considered this, but I've decided I want it to be really obvious that I'm invalidating a record in the set I'm about to move.

neoeno:
:+1:

darokel:
:+1:

PR #59, opened 2015-05-12
IndexFirmJob needs to be passed the firm not just an ID

munckymagik:
Was passing the wrong type! Gave it an numeric id but it needed a firm object. It told me about the problem by raising an error in the firm_serializer. Doh.

`<sssshhhh>`Oh for the days of parameter type checking`</sssshhhh>`


neoeno:
Aha — good catch. Hope I haven't fallen victim to this myself! :+1:

darokel:
:+1:

PR #62, opened 2015-06-03
Cascade from principal

munckymagik:
Changes firm->principal association so that destroying a firm no longer destroys the associated principal. The principal may be linked to other firms and we should not leave them orphaned.

Also changes `Principal` so that associated firms are destroyed when destroying their principal. As the principal is the entity which ties everything together, it doesn't make sense to retain any firms linked to them.

@neoeno wdyt?

### DIFF
--- a/app/models/principal.rb
+++ b/app/models/principal.rb
@@ -7,7 +7,8 @@ class Principal < ActiveRecord::Base
   has_one :firm,
     -> { where(parent_id: nil) },
     primary_key: :fca_number,
-    foreign_key: :fca_number
+    foreign_key: :fca_number,
+    dependent: :destroy
 
   validates :fca_number,
     presence: true,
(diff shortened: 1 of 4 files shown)


app/models/principal.rb
```diff
@@ -7,7 +7,8 @@ class Principal < ActiveRecord::Base
   has_one :firm,
     -> { where(parent_id: nil) },
     primary_key: :fca_number,
-    foreign_key: :fca_number
+    foreign_key: :fca_number,
+    dependent: :destroy
```
munckymagik: The parent firm is already configured to cascade delete subsidiary (trading name) firms, so all we have to do is delete the primary firm associated with the principal to get the effect we want.

neoeno:
seems good :+1:

rnalexander:
:+1: :shipit:

PR #63, opened 2015-06-03
Refactor/gradual rename subs to trading names

munckymagik:
The terminology in the domain model is a bit mixed at the moment. A firm can have one or more "trading names", however in our current schema they are referred to as subsidiaries which is a different thing in the real world.

This PR begins a process of migrating the model to use the same language as we are using in the view, by simply aliasing several key bits so that client code in `rad` looks less confusing.


rnalexander:
:+1:

pmcostello:
:+1:

PR #65, opened 2015-06-09
Firm sorting scopes

munckymagik:
I will probably squash the last rename commit once we're done here.

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -14,6 +14,10 @@ class Firm < ActiveRecord::Base
   scope :registered, -> { where.not(email_address: nil) }
   scope :sorted_by_registered_name, -> { order(:registered_name) }
 
+  def self.most_recently_updated(limit: nil)
+    order(updated_at: :desc).limit(limit)
+  end
+
   has_and_belongs_to_many :in_person_advice_methods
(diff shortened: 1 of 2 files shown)


app/models/firm.rb
```diff
@@ -12,6 +12,11 @@ class Firm < ActiveRecord::Base
   ]
 
   scope :registered, -> { where.not(email_address: nil) }
+  scope :sorted_by_registered_name, -> { order(:registered_name) }
+
+  def self.most_recently_updated(limit: nil)
+    order(updated_at: :desc).limit(limit)
+  end
```
munckymagik: Made the 2nd one a class method to keep Rubocop happy: it was too long as a proc and [this](https://github.com/bbatsov/ruby-style-guide/issues/270) ...

neoeno: There any sense in giving this a default value? (I don't know)

neoeno:
:+1:

rnalexander:
:+1:

app/models/firm.rb
munckymagik: Me neither. Currently I'm seeing the limit as an application side decision, let's see how it goes.

PR #66, opened 2015-06-11
Use Faker to produce unique names.

neoeno:
Sometimes everyone being named Ben Lovell can cause false positives.
It's hard to tell whether the Ben Lovell you have is the one you want.


neoeno:
Clearly this doesn't work like I thought it did.

PR #74, opened 2015-06-25
Fix test which relied on default ordering of db result sets

munckymagik:
We can't really rely on unsorted results being returned in the order
they were put in. So instead we create more records in unsorted order (8
now instead of 2) and simply check they definitely come out sorted from
the scope we're testing.


neoeno:
nice :+1:

benbarnett:
:+1:

PR #75, opened 2015-06-29
Strip out recently used scopes

munckymagik:
These two scopes were introduced for the self_service dashboard feature in RAD, which has now been removed.

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -37,8 +37,6 @@ class Adviser < ActiveRecord::Base
   after_commit :geocode
   after_commit :reindex_old_firm
 
-  scope :most_recently_edited, -> (limit: 3) { order(updated_at: 'DESC').limit(limit) }
-
   def self.on_firms_with_fca_number(fca_number)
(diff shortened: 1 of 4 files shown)


urfolomeus:
With the caveat that I don't have enough context to spot any potential side effects of removing these ... :+1: :)

munckymagik:
OK to give a bit of assurance, they were only ever used on one page in RAD and where taken out in this commit (now in rad master):

* https://github.com/moneyadviceservice/rad/commit/3e0a902ad78184890a33e8648b57fb5d35c9f14c#diff-a945ae7065cd502374b05ded7404783fL11
* https://github.com/moneyadviceservice/rad/commit/3e0a902ad78184890a33e8648b57fb5d35c9f14c#diff-a945ae7065cd502374b05ded7404783fL22

I also have all three RAD codebases checked out into one folder on my machine and grep'd through the lot to see if there were stray instances.

urfolomeus:
Thanks for the context! :)

:100:

darokel:
:+1:

PR #77, opened 2015-07-01
Change business offering percentage fields to booleans.

neoeno:
I decided not to propagate this change to ElasticSearch, because that would mean doing some mucking around with the live server to make all the databases and codebases align. It's possible that we'll do that, but given that this is a priority change and it'll be a while until this is deployed to production — we'll defer it.

`firm_serializer` is responsible for converting booleans to percentages.

### DIFF
--- a/db/migrate/20150630143001_change_percentage_fields_to_boolean_fields_on_firm.rb
+++ b/db/migrate/20150630143001_change_percentage_fields_to_boolean_fields_on_firm.rb
@@ -1,0 +1,35 @@
+class ChangePercentageFieldsToBooleanFieldsOnFirm < ActiveRecord::Migration
+  BUSINESS_OFFERING_ATTRIBUTE = [
+    :retirement_income_products,
+    :pension_transfer,
+    :long_term_care,
+    :equity_release,
+    :inheritance_tax_and_estate_planning,
+    :wills_and_probate,
+    :other]
+
+  def up
+    BUSINESS_OFFERING_ATTRIBUTE.each do |field|
+      # We add the boolean column
+      add_column :firms, "#{field}_flag".to_sym, :boolean, null: false, default: false
+      # Translate the old percentage values to the new boolean column
+      Firm.where("#{field}_percent > ?", 0).update_all("#{field}_flag" => true)
+      # And finally, remove the percentage column
+      remove_column :firms, "#{field}_percent".to_sym, :integer
+    end
+  end
+
+  def down
+    BUSINESS_OFFERING_ATTRIBUTE.each do |field|
+      # As above, but in reverse.
+      # Since we've lost information, we can only migrate back by figuring
+      #   true  => 100%
+      #   false => 0%
+      # Which won't validate since the fields won't add up to 100%, but it
+      # will work in a pinch.
+      add_column :firms, "#{field}_percent".to_sym, :integer
+      Firm.where("#{field}_flag" => true).update_all("#{field}_percent" => 100)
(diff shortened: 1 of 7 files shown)


db/migrate/20150630143001_change_percentage_fields_to_boolean_fields_on_firm.rb
```diff
@@ -0,0 +1,26 @@
...
+    :long_term_care,
+    :equity_release,
+    :inheritance_tax_and_estate_planning,
+    :wills_and_probate,
+    :other]
+
+  def up
+    BUSINESS_OFFERING_ATTRIBUTE.each do |field|
+      add_column :firms, "#{field}_flag".to_sym, :boolean, null: false, default: false
+      Firm.where("#{field}_percent > ?", 0).update_all("#{field}_flag" => true)
```
munckymagik: So this is the key line where we migrate existing percentage based data to the new boolean fields?

neoeno: Yes. At first I did it by pulling out all the data and shoving it back in again but it took ages, and this seemed like a better way. Are you thinking it needs documenting?

db/migrate/20150630143001_change_percentage_fields_to_boolean_fields_on_firm.rb
```diff
@@ -0,0 +1,26 @@
...
+      remove_column :firms, "#{field}_percent".to_sym, :integer
+    end
+  end
+
+  def down
+    BUSINESS_OFFERING_ATTRIBUTE.each do |field|
+      add_column :firms, "#{field}_percent".to_sym, :integer
+      Firm.where("#{field}_flag" => true).update_all("#{field}_percent" => 100)
```
munckymagik: This doesn't back-out cleanly, because we can't get back the original values. What are the pros and cons of this, I wonder?

munckymagik: I mean in terms of, if we actually needed to back out the change from the live system, what state would things be in from the users' perspective?

neoeno: You're right, it doesn't. It seemed like the best thing to do, since the directory only cares about (effectively) truthy or falsey on each field.

The difference would be that in the self-service & signup interface, everything would be either 0 or 100, so actually none of the records would validate based on the pre-change code... which would be a shame, now I think of it.

neoeno: I'd avoid doing it too cleverly to make it all add up to 100% though — but would it be better to simply make it un-downgradeable?

munckymagik: Thinking about this it's OK if we keep the downgrade migration for us as developers. For a live revert, we would restore the backup we take before we ran the migration, then tell Heroku to go back to the last deploy. So I feel this is enough comfort.

db/migrate/20150630143001_change_percentage_fields_to_boolean_fields_on_firm.rb
munckymagik: Yeah, maybe a comment or wrap it in a method so we can give it an obvious marker to show what we're doing.

neoeno: Done

munckymagik:
:+1:

darokel:
:+1:

PR #78, opened 2015-07-06
Move `website_address` attribute from principal to firm.

neoeno:
Firms and trading names should have their own websites, not just use the principal's one.


munckymagik:
:+1:

darokel:
:+1:

PR #80, opened 2015-07-22
Add Remote/face to face advice method logic

neoeno:
This will make more sense when I do the PR on RAD. [EDIT: https://github.com/moneyadviceservice/rad/pull/275]

Also sorry for the rename commit — need to work on my TTD (time-travel driven development)

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -86,8 +86,19 @@ class Firm < ActiveRecord::Base
     allow_blank: true,
     numericality: { only_integer: true }
 
+  validates :in_person_advice_methods,
+    presence: true,
+    if: ->{ remote_or_local_advice == :local }
+
+  validates :other_advice_methods,
+    presence: true,
+    if: ->{ remote_or_local_advice == :remote }
+
   validates *ADVICE_TYPES_ATTRIBUTES,
     inclusion: { in: [true, false] }
+
+  validates :remote_or_local_advice,
+    presence: true
 
   validate do
     unless advice_types.values.any?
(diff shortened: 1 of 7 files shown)


app/models/firm.rb
```diff
@@ -151,6 +165,11 @@ def advice_types
     ADVICE_TYPES_ATTRIBUTES.map { |a| [a, self[a]] }.to_h
   end
 
+  def primary_advice_method
```
tomas-stefano: Where this method is called? Is only on the validation?

tomas-stefano: If it is only on the validation, can we create a custom validator for the case?

neoeno: It's used by RAD too — see here for instance: https://github.com/moneyadviceservice/rad/pull/275/files#diff-6004437aeea98d71ab51826159b8414fR8

Nice idea though — not used custom validators before! Will be sure to in future as big messy validations are one of my least favourite things.

neoeno:
@rnalexander Thanks! I fixed that in a later commit but as you rightly pointed out I can squash it from the history entirely, so I've done that.

@JiggyPete Good idea on the `if:` front. The syntax ended up being slightly different but I've implemented the spirit of it.

JiggyPete:
:+1:

app/models/firm.rb
tomas-stefano: Ok. Then consider next time in creating custom validators. Totally worth it. :)

tomas-stefano:
:+1:

rnalexander:
:+1: :shipit:

PR #82, opened 2015-08-12
Adding an onboarding status check to principal

urfolomeus:
We need to know how far through the onboarding process a principal is.

The aim is to get a Principal to the point where they have at least one firm on the register.

The rules that we are implementing here are as follows:

1. If a Principal has no registered firms (at the moment determined by whether or not they have made it far enough to supply an email address) then they should be prompted to `:complete_a_firm`
2. If they have at least one completed firm then the next check is to see whether or not they should be setting up an adviser. A firm should be prompted to :complete_an_adviser if
    1. They have no remote firms (i.e. ones that primarily offer advice without an in-person visit). This is because remote firms do not need advisers so they will be eligible to be on the register without advisers and therefore the Principal will have a firm on the register
    2. They have no advisers.

Hopefully this makes sense! :)

### DIFF
--- a/app/models/principal.rb
+++ b/app/models/principal.rb
@@ -71,7 +71,31 @@ class Principal < ActiveRecord::Base
     "#{first_name} #{last_name}"
   end
 
+  def next_onboarding_action
+    registered_firms = main_firm_with_trading_names.registered
+
+    if registered_firms.empty?
+      :complete_a_firm
+    elsif needs_advisers?(registered_firms)
+      :complete_an_adviser
+    else
+      :onboarded
+    end
+  end
+
   private
+
+  def needs_advisers?(firm_list)
+     no_remote_firms?(firm_list) && no_advisers_for?(firm_list)
+  end
+
+  def no_remote_firms?(firm_list)
+    firm_list.none? { |f| f.primary_advice_method == :remote }
+  end
+
+  def no_advisers_for?(firm_list)
+    firm_list.none? { |f| f.advisers.present? }
+  end
 
   def find_subsidiary(subsidiary)
     firm.subsidiaries.find_or_initialize_by(
--- a/app/models/principal.rb
+++ b/app/models/principal.rb
@@ -74,11 +74,12 @@ class Principal < ActiveRecord::Base
   def next_onboarding_action
     registered_firms = main_firm_with_trading_names.registered
 
-    if registered_firms.empty?
+    case
+    when registered_firms.empty?
       :complete_a_firm
-    elsif any_remote_firms?(registered_firms)
+    when any_remote_firms?(registered_firms)
       :onboarded
-    elsif needs_advisers?(registered_firms)
+    when needs_advisers?(registered_firms)
       :complete_an_adviser
     else
       :onboarded
--- a/app/models/principal.rb
+++ b/app/models/principal.rb
@@ -76,6 +76,8 @@ class Principal < ActiveRecord::Base
 
     if registered_firms.empty?
       :complete_a_firm
+    elsif any_remote_firms?(registered_firms)
+      :onboarded
     elsif needs_advisers?(registered_firms)
       :complete_an_adviser
     else
@@ -85,16 +87,12 @@ class Principal < ActiveRecord::Base
 
   private
 
+  def any_remote_firms?(firm_list)
+    firm_list.any? { |f| f.primary_advice_method == :remote }
+  end
+
   def needs_advisers?(firm_list)
-     no_remote_firms?(firm_list) && no_advisers_for?(firm_list)
-  end
-
-  def no_remote_firms?(firm_list)
-    firm_list.none? { |f| f.primary_advice_method == :remote }
-  end
-
-  def no_advisers_for?(firm_list)
-    firm_list.none? { |f| f.advisers.present? }
+     firm_list.none? { |f| f.advisers.present? }
   end
 
   def find_subsidiary(subsidiary)
--- a/spec/models/principal_spec.rb
+++ b/spec/models/principal_spec.rb
@@ -204,10 +204,11 @@ RSpec.describe Principal do
   end
 
   describe '#next_onboarding_action' do
-    context 'when principal has no parent firm' do
+    context 'when principal has no firms or trading names' do
       before :each do
         principal.firm.destroy
         principal.reload
+        expect(principal.main_firm_with_trading_names).to be_empty
       end
 
       it 'returns :complete_a_firm' do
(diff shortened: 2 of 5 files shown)


spec/models/principal_spec.rb
```diff
@@ -202,4 +202,325 @@
...
+      end
+
+      it 'returns :complete_a_firm' do
+        expect(principal.next_onboarding_action).to eql(:complete_a_firm)
+      end
+    end
+
+    context 'when principal only has a parent firm' do
+      before :each do
+        expect(principal.firm).not_to be_nil
```
JiggyPete: There are a lot of expectations in before blocks which seems weird.

Are these pre-condition checks to protect against factories/src changing in the future or perhaps to gain confidence in establishing what factory girl and the src is doing?

app/models/principal.rb
```diff
@@ -71,8 +71,32 @@ def full_name
...
+      :complete_an_adviser
+    else
+      :onboarded
+    end
+  end
+
   private
 
+  def needs_advisers?(firm_list)
+     no_remote_firms?(firm_list) && no_advisers_for?(firm_list)
```
JiggyPete: Knowing the logic for this it still took me a good while to get my head around. The negative `no_` aspect, flipping what could be a positive boolean expression, trips me up in my thinking. Perhaps I'm alone in this. Could it be expressed more positively?

spec/models/principal_spec.rb
mconnell: There's a fair bit of setup so I think they were mostly a sanity check, but they effectively also help spot future changes.

app/models/principal.rb
munckymagik: I agree with Pete. We need advisers if there **are any** firms that do offer face-to-face advice (referred to as `in_person_advice` in `firm.rb`). I really think we should positively check for the type that we know does need advisers.

```
any_in_person_advice_firms? && no_advisers_for?(firm_list)
```

urfolomeus: We had this originally but @neoeno _made_ us change it ;)

app/models/principal.rb
```diff
@@ -71,8 +71,32 @@ def full_name
...
+  end
+
+  def no_remote_firms?(firm_list)
+    firm_list.none? { |f| f.primary_advice_method == :remote }
+  end
+
+  def no_advisers_for?(firm_list)
+    firm_list.none? { |f| f.advisers.present? }
```
munckymagik: What do the queries look like behind these two methods? How frequently will they be invoked?

urfolomeus: We'll not really know until we've worked out where and when it gets used in rad. Once we've done that we can work on optimising for performance.

spec/models/principal_spec.rb
```diff
@@ -202,4 +202,325 @@
       expect(Firm.where(id: firm.id)).to be_empty
     end
   end
+
+  describe '#next_onboarding_action' do
+    context 'when principal has no parent firm' do
+      before :each do
+        principal.firm.destroy
```
munckymagik: A principal with a nil/destroyed parent firm is not a real state in the system at the moment. The `after_create  :associate_firm` handler in principal.rb guarantees there is a skeleton firm from the moment it's created.

So I'm not sure if we should be handling this scenario like it's a valid state. If this were to happen it means something wrong has happened or the system has stopped behaving consistently.

app/models/principal.rb
urfolomeus: On a more serious note this logic is, by it's nature, confusing. That is why we've had several long conversations on it, why I have a spreadsheet listing all of the state permutations on my desktop, and why we have close to 400 lines of tests around this. We landed at this implementation being the most descriptive. Happy to change it if a better solution is found though.

We did stick for a while at the negative naming, but the situation we are coding for is this:

* If the principal has _at least once_ **remote** firm then they are `:onboarded` as remote firms are able to be added to the register as long as the firm is complete.
* If they have _no_ **remote** firms then we check to see if they need advisers

We did originally have the method to check for "not remote" being "for in-person", but it was actually misleading as the mind automatically goes to the version that Dan has above, which is actually incorrect. It needs to be _all_ in_person, not _any_ in_person. :)

spec/models/principal_spec.rb
urfolomeus: Yeah, we did wonder about this. Ultimately we included it for completeness. There are a few other permutations that are (currently) impossible. Happy to remove them if it's confusing though.

spec/models/principal_spec.rb
```diff
@@ -202,4 +202,325 @@
...
+      end
+    end
+
+    context 'when principal only has a parent firm' do
+      before :each do
+        expect(principal.firm).not_to be_nil
+        expect(Firm.where(fca_number: principal.fca_number).count).to eql(1)
+      end
+
+      context 'and the parent firm has no email address' do
```
munckymagik: This is confusing. The question we should be asking here is: do we consider the firm to be fully registered yet? Email address being nil is misleading and a bit leaky.

The fact that firms consider themselves registered if their `email_address` has been filled out is a shameful truth that firms would rather keep secret. They don't want us telling other bits of the system or relying on this knowledge. It's embarrassing for firms. Firms would rather you ask "are you `registered?`"

Seriously though, I feel it's time to encapsulate this logic in one place.

spec/models/principal_spec.rb
```diff
@@ -202,4 +202,325 @@
...
+          end
+        end
+
+        context 'and the firm primarily gives in-person advice' do
+          before :each do
+            principal.firm.in_person_advice_methods << create(:in_person_advice_method)
+            principal.reload
+          end
+
+          context 'and has at least one in-person adviser' do
```
munckymagik: There is no such thing as an "in-person adviser", can we remove the "in-person" prefix or clarify what we mean here?

spec/models/principal_spec.rb
urfolomeus: I agree with this. It's something that we struggled with when writing these tests.

spec/models/principal_spec.rb
urfolomeus: Yes. Will do.

app/models/principal.rb
```diff
@@ -71,8 +71,30 @@ def full_name
     "#{first_name} #{last_name}"
   end
 
+  def next_onboarding_action
+    registered_firms = main_firm_with_trading_names.registered
+
+    if registered_firms.empty?
+      :complete_a_firm
+    elsif any_remote_firms?(registered_firms)
+      :onboarded
```
munckymagik: Yes, I like this. This short-circuit makes this much easier to understand what's really happening. :100:

JiggyPete: :+1:

urfolomeus:
ok @JiggyPete @munckymagik I think I've addressed all concerns now. Can you review and comment please? Ta :)

munckymagik:
:+1:

JiggyPete:
:+1:

darokel:
:+1:

PR #84, opened 2015-08-14
[DEV APPROVED] Basics of multiple offices for firms

munckymagik:
This is the first PR for this feature and includes just the basics required to implement the firms offices index page in RAD.

**Note**: there are no validations yet as the new/edit forms have not been implemented in RAD and I'm driving implementation from there.

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -25,7 +25,7 @@ class Firm < ActiveRecord::Base
   belongs_to :parent, class_name: 'Firm'
 
   has_many :advisers, dependent: :destroy
-  has_many :offices
+  has_many :offices, -> { order created_at: :asc }
   has_many :subsidiaries, class_name: 'Firm', foreign_key: :parent_id, dependent: :destroy
   has_many :trading_names, class_name: 'Firm', foreign_key: :parent_id, dependent: :destroy
   has_many :qualifications, -> { reorder('').uniq }, through: :advisers
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -25,6 +25,7 @@ class Firm < ActiveRecord::Base
   belongs_to :parent, class_name: 'Firm'
 
   has_many :advisers, dependent: :destroy
+  has_many :offices
   has_many :subsidiaries, class_name: 'Firm', foreign_key: :parent_id, dependent: :destroy
   has_many :trading_names, class_name: 'Firm', foreign_key: :parent_id, dependent: :destroy
   has_many :qualifications, -> { reorder('').uniq }, through: :advisers
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -178,6 +178,10 @@ class Firm < ActiveRecord::Base
     infer_primary_advice_method
   end
 
+  def main_office
+    offices.first
+  end
+
   private
 
   def delete_elastic_search_entry
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -1,0 +1,4 @@
+class Office < ActiveRecord::Base
+  belongs_to :firm
+end
+
--- a/spec/models/firm_spec.rb
+++ b/spec/models/firm_spec.rb
@@ -82,16 +82,16 @@ RSpec.describe Firm do
     let(:firm) { create(:firm) }
     let!(:unsorted_offices) do
       [
-        FactoryGirl.create(:office, firm: firm, address_line_one: 'd', created_at: Time.zone.now),
-        FactoryGirl.create(:office, firm: firm, address_line_one: 'b', created_at: 2.days.ago),
-        FactoryGirl.create(:office, firm: firm, address_line_one: 'a', created_at: 3.days.ago),
-        FactoryGirl.create(:office, firm: firm, address_line_one: 'c', created_at: 1.day.ago)
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'fourth', created_at: Time.zone.now),
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'second', created_at: 2.days.ago),
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'first',  created_at: 3.days.ago),
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'third',  created_at: 1.day.ago)
       ]
     end
 
     describe 'default sort order' do
       subject { firm.offices.map(&:address_line_one) }
-      it { is_expected.to eq(%w{a b c d}) }
+      it { is_expected.to eq(%w{first second third fourth}) }
     end
   end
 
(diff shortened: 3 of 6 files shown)


spec/models/firm_spec.rb
```diff
@@ -66,6 +66,37 @@
...
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'd', created_at: Time.zone.now),
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'b', created_at: 2.days.ago),
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'a', created_at: 3.days.ago),
+        FactoryGirl.create(:office, firm: firm, address_line_one: 'c', created_at: 1.day.ago)
+      ]
+    end
+
+    describe 'default sort order' do
+      subject { firm.offices.map(&:address_line_one) }
+      it { is_expected.to eq(%w{a b c d}) }
```
JiggyPete: Just looking at the test it reads like the default sort order is based on `:address_line_one`

I'd suggest changing the values to `a => first, b => second, c=> third, d => fourth` as that would have the same meaning (as in this is the order offices should be returned in), however these values wouldn't be ordered alphabetically as ordering is by `:created_at`.

app/models/firm.rb
```diff
@@ -25,6 +25,7 @@ class Firm < ActiveRecord::Base
   belongs_to :parent, class_name: 'Firm'
 
   has_many :advisers, dependent: :destroy
+  has_many :offices, -> { order created_at: :asc }
```
urfolomeus: As offices are specific to firms, should we not be removing them i the firm is deleted?

app/models/office.rb
```diff
@@ -0,0 +1,4 @@
+class Office < ActiveRecord::Base
+  belongs_to :firm
+end
+
```
urfolomeus: You have a few `null: false` fields in the migration, but no validations.

spec/models/firm_spec.rb
```diff
@@ -66,6 +66,37 @@
...
+    let(:firm) { create(:firm) }
+    subject { firm.main_office }
+
+    context 'when the firm has no offices' do
+      it { is_expected.to be_nil }
+    end
+
+    context 'when the firm has offices' do
+      before { FactoryGirl.create_list(:office, 3, firm: firm) }
+      it { is_expected.to eq(firm.offices[0]).and(eq(firm.offices.first)) }
```
urfolomeus: Is testing `#[]` and `.first` in the same test not testing the framework rather than the app's behaviour?

app/models/office.rb
JiggyPete: validations will come later

urfolomeus: why? :)

spec/models/firm_spec.rb
munckymagik: That's a very good idea.

app/models/firm.rb
munckymagik: Yes exactly right. That's on my todo list.

spec/models/firm_spec.rb
JiggyPete: it's testing that the firm has a sort order and this method returns the first result

app/models/office.rb
JiggyPete: says so at the top of the PR

app/models/firm.rb
munckymagik: And also it's listed in the acceptance criteria on the card: https://trello.com/c/5ab7WP1I

app/models/office.rb
munckymagik: Yes, see description up top :)

spec/models/firm_spec.rb
urfolomeus: Oh, do `firm.offices[0]` and `firm.offices.first` return different results from one another?

app/models/firm.rb
urfolomeus: Oops, sorry I should have read that :blush:

app/models/office.rb
urfolomeus: Sorry, forgot to read that bit >.<

spec/models/firm_spec.rb
munckymagik: In a way yes. But, more what I'm testing is my understanding of how the framework behaves is correct (and continues to be) so these facts can be relied upon.

urfolomeus:
:+1:

spec/models/firm_spec.rb
munckymagik: Remember `.first` and it's habit of doing an `ORDER BY id` unless you specify something explicit? Well there is an explicit order on the `.offices` collection and I want to be sure I'm getting that.

JiggyPete: nope. I misunderstood...one of them could go

JiggyPete: fair enough @munckymagik 
A test providing confidence is a good thing

JiggyPete:
:+1:

app/models/firm.rb
```diff
@@ -173,6 +174,10 @@ def primary_advice_method
     infer_primary_advice_method
   end
 
+  def main_office
+    offices.first
```
Tr4pSt3R: is there no business specification of what ```main_office``` entails?

munckymagik: Hi Jones. What do you mean by 'business specification'?

Tr4pSt3R: The definition of ```main office``` should be more than just the first office in the list. If that's the case then there's no need in even creating a method for specifying it. I hope it makes sense (it does in my head).

paulanthonywilson: I think I know what @Tr4pSt3R means.

There's no special meaning to "main office". The consumer facing side still assumes that there each firm has a single office, and that won't change in the short term. So we have to support just displaying one office, consistently.

We discussed this with the PM, and it's fine for  this _main office_ to simply be the first office added.

Tr4pSt3R: @paulanthonywilson @munckymagik okay that's fine

munckymagik: Umm I'm confused. Does this help? ...

https://github.com/moneyadviceservice/rad/blob/multiple_offices/app/views/self_service/offices/_offices_table.html.erb#L32

and

screen shot 2015-08-14 at 14 14 07

munckymagik: The rule I'm following is: the offices are sorted by `created_at`. The first office created is considered the "main office". This means we can treat that record as special, e.g. prevent it being deleted through the UI and so on.

munckymagik:
@Tr4pSt3R could I get a thumb?

asmega:
:+1:

PR #87, opened 2015-08-20
[DEV APPROVED] Multiple offices (part deux)

munckymagik:
This set of changes adds validation rules and behaviour, driven by the need to implement add/edit pages in RAD.

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -22,9 +22,13 @@ class Office < ActiveRecord::Base
     presence: true,
     format: { with: /\A[A-Z\d]{1,4} [A-Z\d]{1,3}\z/ }
 
-  validates :address_town, presence: true
+  validates :address_town,
+    presence: true,
+    length: { maximum: 100 }
 
-  validates :address_county, presence: true
+  validates :address_county,
+    presence: true,
+    length: { maximum: 100 }
 
   validates :disabled_access, inclusion: { in: [true, false] }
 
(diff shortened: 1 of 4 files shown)


spec/support/field_length_validation_helpers.rb
```diff
@@ -0,0 +1,10 @@
+module FieldLengthValidationHelpers
+  # USAGE: expect_length_of(subject, :field_name, 50).to be_valid
+  def expect_length_of(subject, field, length, fill_char: 'x')
+    num_chars_to_fill = (length - subject.send(field).length)
+    new_value = subject.send(field) + (fill_char * num_chars_to_fill)
+    subject.send("#{field}=", new_value)
+
+    expect(subject)
+  end
```
neoeno: neat

neoeno:
:+1:

spec/support/field_length_validation_helpers.rb
munckymagik: thanks

app/models/office.rb
```diff
@@ -1,4 +1,52 @@
 class Office < ActiveRecord::Base
   belongs_to :firm
+
+  validates :email_address,
+    presence: true,
+    length: { maximum: 50 },
+    format: { with: /.+@.+\..+/ }
```
benbarnett: but what if I ***AM*** a TLD?

benbarnett: :trollface: btw

benbarnett:
:+1:

app/models/office.rb
munckymagik: :) Yes exactly. It stops subversives like me putting in dan@localhost when testing.

PR #88, opened 2015-08-21
Simplify and optimise languages validation

munckymagik:
Now caches the available languages at class load time and uses the Set
data structure to simplify and optimise validation of selected
languages.

The existing implementation involved a lot of nested loops which meant
its runtime would be sensitive to changes in the number of languages
available and selected. Our rough estimate of its running time
complexity was O(n^3).

The new approach guarantees the running time will not vary so much. It
should now be O(n).

@neoeno

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -105,7 +105,7 @@ class Firm < ActiveRecord::Base
     presence: true
 
   validate :languages do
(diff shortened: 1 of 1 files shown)


neoeno:
:+1:

asmega:
:+1:

PR #89, opened 2015-08-24
Cascade delete offices from firms

munckymagik:
Enhances the Firms -> Offices association to delete offices when the owning firm is deleted.


JiggyPete:
:+1:

asmega:
:+1:

PR #90, opened 2015-08-24
[DEV APPROVED] adding principal.onboarded?

JiggyPete:
Rather than propogate the states within the :next_onboarding_action around the rad codebase it would be nicer to be able to ask a direct question of the principal. Have they been onboarded?

### DIFF
--- a/spec/factories/firm.rb
+++ b/spec/factories/firm.rb
@@ -25,6 +25,15 @@ FactoryGirl.define do
     longitude { Faker::Address.longitude.to_f.round(6) }
     status :independent
 
+    factory :onboarded_firm do
+      parent factory: Firm
+    end
+
+    factory :not_onboarded_firm do
+      parent factory: Firm
+      email_address nil
+    end
+
     factory :trading_name, aliases: [:subsidiary] do
       parent factory: Firm
     end
--- a/spec/models/principal_spec.rb
+++ b/spec/models/principal_spec.rb
@@ -203,6 +203,30 @@ RSpec.describe Principal do
     end
   end
 
+  describe '#onboarded?' do
+    context 'when a principal has been onboarded' do
+      before :each do
+        principal.firm.update_attributes(FactoryGirl.attributes_for :onboarded_firm, fca_number: principal.fca_number)
+        create(:adviser, firm: principal.firm)
+      end
+
+      it 'returns true' do
+        expect(principal).to be_onboarded
+      end
+    end
+
+    context 'when a principal has not been onboarded' do
+      before :each do
+        attrs = FactoryGirl.attributes_for :not_onboarded_firm, fca_number: principal.fca_number
+        principal.firm.update_attributes(attrs)
+      end
+
+      it 'returns false' do
+        expect(principal).not_to be_onboarded
+      end
+    end
+  end
+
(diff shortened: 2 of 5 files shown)


spec/models/principal_spec.rb
```diff
@@ -203,6 +203,25 @@
     end
   end
 
+  describe '#onboarded?' do
+    context 'when a principal has been onboarded' do
+      before :each do
+        principal.firm.update_column(:email_address, 'acme@example.com')
```
munckymagik: It's a shame this email address thing is still spreading around the codebase. I wonder what we can do to encapsulate it better?

munckymagik: Plus, simply setting the email address doesn't really create an onboarded firm. Is there something missing in the logic that decides what onboarded means?

Maybe for this test what we really should do is mock `next_onboarding_action` to return known states.

JiggyPete: I'll try and wrap this up in a factory. A factory isn't really the best place for encapsulating logic however it will be reusable.

Re: email address:
Currently the logic is if the firm is registered AND if they are remote 
OR
firm is registered AND has an adviser.

For the firm to be registered they need to have been valid at the time of registering.

re: tests
If we were to mock a call on the same object that'd be a smell that we need to change the design and push the :next_onboarded_action somewhere else. Not sure where though. 
Given the amount of setup I'm happy with it living on the principal and the :onboarded? tests unless there is somewhere clearer for the logic to live?

JiggyPete:
Hi @munckymagik, I've added in 3 firm factories, `:registered_firm` `:onboarded_firm` and `:not_onboarded_firm`. Hope this helps keep the ugliness in one place, thoughts?

spec/factories/firm.rb
```diff
@@ -35,6 +35,19 @@
...
+
+    factory :onboarded_firm do
+      parent factory: :registered_firm
+    end
+
+    factory :not_onboarded_firm do
+      parent factory: :firm
+      email_address nil
+    end
+
```
munckymagik: Ok, this feels better. It stops the spread of implementation details. I think this factory is due a rethink at some point though.

JiggyPete: Yeah, as you said earlier we need a nicer way of handling `:registered`.

munckymagik:
:+1:

asmega:
:+1: 

although one point outside the scope of this PR

- `next_onboarding_action` I would of thought this method would return falsey if there were no more onboarding actions.

JiggyPete:
@asmega the `:next_onboarding_action` method does look a bit strange. The current implementation was the clearest solution we could find. 

I think you getting at is the `else` case? Or would you replace both instances of `:onboarded` with falsey?

asmega:
I'm fine with the case statement and else statement. This looks odd to me `next_onboarding_action == :onboarded`

Unless `:onboarded` is a legitimate next onboarding action in which case the code is perfectly fine.

JiggyPete:
Ahh. Currently we have that code in a branch of rad, which is why we are trying to keep the nasty within principal with this PR.

We probably would have had falsey/nil, however as there are two places in the case for this we were trying to be more explicit.

PR #91, opened 2015-08-24
Refactor firm factory to use traits and tidy up the the firm_spec

munckymagik:
It has reached a point where we really need to be able to combine traits when constructing Firms using FactoryGirl. Refactored the factory to make that possible. Tests ran against

* mas-rad_core
* rad (master)
* rad_consumer (dev/upgrade_mrc)

To ensure nothing has been broken.

Also tidied tests related to deleting firms to take out unnecessary calls to after_commit hooks etc. prompted by @JiggyPete's comment in #89.

### DIFF
--- a/spec/factories/firm.rb
+++ b/spec/factories/firm.rb
@@ -30,10 +30,8 @@ FactoryGirl.define do
     longitude { Faker::Address.longitude.to_f.round(6) }
     status :independent
 
-    factory :trading_name do
+    factory :trading_name, aliases: [:subsidiary] do
       parent factory: Firm
-
-      factory :subsidiary # alias for trading_name
     end
 
     factory :firm_with_advisers, traits: [:with_advisers]
--- a/spec/factories/firm.rb
+++ b/spec/factories/firm.rb
@@ -30,18 +30,21 @@ FactoryGirl.define do
     longitude { Faker::Address.longitude.to_f.round(6) }
     status :independent
 
-    factory :firm_with_remote_advice do
-      other_advice_methods { create_list(:other_advice_method, rand(1..3)) }
-      in_person_advice_methods []
-    end
-
     factory :trading_name do
       parent factory: Firm
 
-      factory :subsidiary
+      factory :subsidiary # alias for trading_name
     end
 
-    factory :firm_with_no_business_split do
+    factory :firm_with_advisers, traits: [:with_advisers]
+    factory :firm_with_offices, traits: [:with_offices]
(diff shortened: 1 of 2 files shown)


asmega:
:+1:

spec/factories/firm.rb
```diff
@@ -30,18 +30,19 @@
     longitude { Faker::Address.longitude.to_f.round(6) }
     status :independent
 
-    factory :firm_with_remote_advice do
-      other_advice_methods { create_list(:other_advice_method, rand(1..3)) }
-      in_person_advice_methods []
-    end
-
-    factory :trading_name do
+    factory :trading_name, aliases: [:subsidiary] do
```
neoeno: nice, didn't know you could do that

spec/factories/firm.rb
```diff
@@ -30,18 +30,19 @@
...
     end
 
-    factory :firm_with_no_business_split do
+    factory :firm_with_advisers, traits: [:with_advisers]
+    factory :firm_with_offices, traits: [:with_offices]
+    factory :firm_with_principal, traits: [:with_principal]
+    factory :firm_with_no_business_split, traits: [:with_no_business_split]
+    factory :firm_with_remote_advice, traits: [:with_remote_advice]
+    factory :firm_with_subsidiaries, traits: [:with_trading_names]
+    factory :firm_with_trading_names, traits: [:with_trading_names]
```
neoeno: sweeeeeeeet

spec/models/firm_spec.rb
```diff
@@ -461,18 +461,8 @@
       it 'cascades destroy to advisers' do
         adviser = firm.advisers.first
         firm.destroy
-        firm.run_callbacks(:commit)
-        adviser.run_callbacks(:commit)
         expect(Adviser.where(id: adviser.id)).to be_empty
```
neoeno: These unnecessary now?

neoeno:
:+1:

spec/models/firm_spec.rb
munckymagik: I believe so. In this test we're not testing side-effects from after_commit hooks we're testing just the cascading delete. The after_commit hooks are tested on line 515. Unless there's something we've missed?

spec/factories/firm.rb
munckymagik: Yes very, isn't it.

munckymagik: Cake and eating ...

PR #92, opened 2015-08-25
[DEV APPROVED] Amend language list (add UK minority languages)

neoeno:
Involved:

* Extracting languages to common module in lib.
* Using ISO 639-3 over 639-1 codes, since we now use some languages without ISO 639-1 codes.
* Remove `Firm.available_languages` since this is a convenient time to refer to it directly in RAD.


paulanthonywilson:
Much disappoint for The Cornish, but :+1:

asmega:
:+1:

PR #93, opened 2015-08-27
Add AdviserResult objects to a FirmResult

mconnell:
When a search is retrieved from elastic search, it is transformed from a JSON result into a `SearchResult` object, which in turn has `FirmResult` objects. Each firm also has a collection of advisers, which are returned in the JSON, but currently not available via the ruby objects. This PR adds in an `AdviserResult` object and adds `FirmResult#advisers`

### DIFF
--- a/lib/mas/adviser_result.rb
+++ b/lib/mas/adviser_result.rb
@@ -1,0 +1,12 @@
+class AdviserResult
+  Location = Struct.new(:latitude, :longitude)
+
+  attr_reader :id, :name, :range, :location
+
+  def initialize(data)
+    @id       = data['_id']
+    @name     = data['name']
+    @range    = data['range']
+    @location = Location.new data['location']['lat'], data['location']['lon']
+  end
+end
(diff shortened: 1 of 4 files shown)


benbarnett:
:+1:

lib/mas/adviser_result.rb
```diff
@@ -0,0 +1,12 @@
+class AdviserResult
+  Location = Struct.new(:latitude, :longitude)
```
paulanthonywilson: I like this

spec/lib/mas/firm_result_spec.rb
```diff
@@ -128,6 +128,17 @@
...
+    describe '#advisers' do
+      it 'returns an array containing the advisers' do
+        expect(subject.advisers).to be_an(Array)
+        expect(subject.advisers.length).to eq(1)
+      end
+
+      it 'returns AdviserResult objects' do
+        expect(subject.advisers.first).to be_an(AdviserResult)
+      end
+    end
```
neoeno: for future reference we have rspec collection matchers in this codebase. not worth amending though unless you really really want to

neoeno:
:+1:

PR #95, opened 2015-08-31
Update Offices

JiggyPete:
This PR is to migrate the address fields out of `firm.rb` and into `office.rb`.

FYI a firm can have multiple offices.

Along the way we also renamed after_commit hooks from `.geocode` to `.geocode_and_reindex` to more clearly explain what the method does.

### DIFF
--- a/spec/jobs/index_firm_job_spec.rb
+++ b/spec/jobs/index_firm_job_spec.rb
@@ -1,14 +1,30 @@
 RSpec.describe IndexFirmJob, '#perform' do
-  let(:firm) { create(:adviser).firm }
   let(:repository) { instance_double(FirmRepository) }
+  let(:firm) { Firm.new }
 
   before do
     allow(FirmRepository).to receive(:new).and_return(repository)
   end
 
-  it 'delegates to the firm repository' do
-    expect(repository).to receive(:store).with(firm)
+  context 'when firm is publishable' do
+    before do
+      allow(firm).to receive(:publishable?).and_return(true)
+    end
 
-    described_class.new.perform(firm)
+    it 'delegates to the firm repository' do
+      expect(repository).to receive(:store).with(firm)
+      described_class.new.perform(firm)
+    end
+  end
+
+  context 'when firm is not publishable' do
+    before do
+      allow(firm).to receive(:publishable?).and_return(false)
+    end
+
+    it 'does not delegates to the firm repository' do
+      expect(repository).not_to receive(:store).with(firm)
+      described_class.new.perform(firm)
+    end
   end
 end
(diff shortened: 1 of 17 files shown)


spec/jobs/index_firm_job_spec.rb
```diff
@@ -1,14 +1,30 @@
...
+      described_class.new.perform(firm)
+    end
+  end
+
+  context 'when firm is not publishable' do
+    before do
+      allow(firm).to receive(:publishable?).and_return(false)
+    end
+
+    it 'delegates to the firm repository' do
```
urfolomeus: Should this be "does not delegate"?

spec/models/adviser_spec.rb
```diff
@@ -142,35 +144,18 @@
         adviser.run_callbacks(:commit)
       end
 
-      def queue_contains_a_job_for(job_class)
+      def aqueue_contains_a_job_for(job_class)
```
urfolomeus: Typo? Should this be `a_queue_contains_a_job_for` or just `queue_contains_a_job_for`?

spec/models/adviser_spec.rb
```diff
@@ -186,19 +171,19 @@ def save_with_commit_callback(model)
     end
 
     context 'when the firm has changed' do
-      it 'triggers reindexing of the adviser and new firm' do
-        expect(GeocodeAdviserJob).to receive(:perform_later).with(subject)
+      it 'triggers reindexing of the original firm the first time' do
         subject.firm = receiving_firm
         save_with_commit_callback(subject)
+        expect(queue_contains_a_job_for(IndexFirmJob)).to be_truthy
```
urfolomeus: Does `queue_contains_a_job_for` exist elsewhere? Is it different to `aqueue_contains_a_job_for`?

spec/jobs/index_firm_job_spec.rb
JiggyPete: Good catch!

spec/models/adviser_spec.rb
JiggyPete: ahh, oversight when refactoring. The method now lives in queue_spec_helper.rb. Will tidy up.

spec/models/adviser_spec.rb
JiggyPete: see above

urfolomeus: Ah, it looks like `queue_contains_a_job_for` is in the `QueueSpecHelper` below and it does have exactly the same implementation, so this can go. :)

urfolomeus: Sorry, didn't see your reply until after I'd added the above :)

urfolomeus:
Looks good. :+1:

rnalexander:
:+1: :shipit:

PR #96, opened 2015-09-01
Fix validations on Firm

neoeno:
* Add some into the fields order method
* Make the #minimum_fixed_fee validations more comprehensive, and
  prevent 500 errors when an integer too large is supplied


mconnell:
:+1:

asmega:
:+1:

PR #99, opened 2015-09-08
Fix the firm factory and add a confidence test

munckymagik:
Following the "3 strikes and you're out rule", I felt it was time to do something about this firm factory: several times now we have discovered that the factories are not producing valid fixtures, in which case how can we trust our tests?

# Goals

1. Correct the factory definitions that do not produce valid models
2. Think of a way to sanity check the factories going forward

# Outstanding work

I decided not to fix the `:firm_with_principal` fixture at this point. We need to upgrade the `rad_consumer` project from using 0.0.74 of `mas-rad_core` to 0.0.78, there are a lot of broken tests there right now I don't want to add to that work until we can see the size of it clearly (this upgrade is my actual task, I just wanted to sort this out and PR it first)

### DIFF
--- a/spec/models/firm_factory_spec.rb
+++ b/spec/models/firm_factory_spec.rb
@@ -21,6 +21,7 @@ RSpec.describe 'Firm factory' do
       it { expect(subject.principal).not_to be_present }
       it { is_expected.to have(:no).offices }
       it { is_expected.to have(:no).advisers }
+      it { is_expected.to have(:no).trading_names }
     end
   end
 
@@ -38,6 +39,7 @@ RSpec.describe 'Firm factory' do
       it { expect(subject.principal).not_to be_present }
       it { is_expected.to have(1).offices }
       it { is_expected.to have(1).advisers }
+      it { is_expected.to have(:no).trading_names }
     end
   end
 
@@ -55,6 +57,7 @@ RSpec.describe 'Firm factory' do
       it { expect(subject.principal).not_to be_present }
       it { is_expected.to have(:no).offices }
       it { is_expected.to have(:no).advisers }
+      it { is_expected.to have(:no).trading_names }
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/models/firm_factory_spec.rb
```diff
@@ -0,0 +1,189 @@
+RSpec.describe 'Firm factory' do
+  subject do
+    begin
+      FactoryGirl.create(factory)
+    rescue ActiveRecord::RecordInvalid
```
asmega: not sure about this. don't you want to be explicit in your test on which it is?

munckymagik: @asmega this is a really good point. I deliberated about this a little. The reason I chose to do it this way is, we can test what really happens rather than forcing it one way or another. At the point I implemented this replacing the default subject `create` with a `build` showed up no breakages.

munckymagik: ... and the explicitness is then provided by doing a `expect(subject).to be_persisted`.

munckymagik: Will add a comment to clarify this decision.

munckymagik: @asmega see 1c4ca57. wdyt?

asmega: the comment makes it better. i guess i don't know enough but I would of thought that if the create fails that would sufficient and there would be no need to initialise the object and test that instead. if your apps behaving like that under the hood then i guess it's fine

munckymagik: @asmega some fixtures need to be created in an invalid state and therefore cannot be persisted. So yes, we do want these tests to fall back to `build`.

asmega:
:+1:

urfolomeus:
Actually I've come round to this. It explicitly states our understanding of the various "states" that a firm can be in.

And yes, I did say _states_

https://media.giphy.com/media/lfbxexWy71b6U/giphy.gif

urfolomeus:
:+1:

munckymagik:
Thanks guys

PR #100, opened 2015-09-08
Modify :firm factory so it creates publishable Firms by default

munckymagik:
Before we implemented multiple offices creating a Firm using the `:firm` factory meant you could assume it would be indexed into ElasticSearch. That got broken because the default `:firm` factory created no offices and to be 'publishable' Firms need at least one. This broke many tests in rad_consumer where everything needs to be accessed via ElasticSearch.

Therefore we now create at least 1 office by default, but you can trigger creation of more by setting the `offices_count` transient.

Note all tests in current rad master pass running against this branch.

### DIFF
--- a/spec/factories/firm.rb
+++ b/spec/factories/firm.rb
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   sequence(:registered_name) { |n| "Financial Advice #{n} Ltd." }
 
-  factory :firm do
+  factory :firm, aliases: [:publishable_firm] do
     fca_number
     registered_name
     email_address { Faker::Internet.email }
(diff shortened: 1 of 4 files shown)


spec/factories/firm.rb
```diff
@@ -25,11 +25,19 @@
...
+    transient do
       offices_count 1
     end
 
+    after(:create) do |firm, evaluator|
+      create_list(:office, evaluator.offices_count, firm: firm)
+      firm.reload
+    end
+
+    factory :onboarded_firm, traits: [:with_advisers] do
```
JiggyPete: I don't want this to grown more arms and legs, however :publishable_firm and :onboarded_firm are theoretically the same thing and if that makes sense we could bin one of them.

JiggyPete:
:+1:

spec/factories/firm.rb
munckymagik: Agreed. We need to fix `Firm#publishable?` then they really will be the same. However at this point there is a difference. Will probably propose a change to do this later. This at least is the minimum change needed to get rad_consumer upgraded to use the same version of this library as rad is using.

asmega:
:+1:

munckymagik:
Thanks guys.

PR #101, opened 2015-09-09
[DEV APPROVED] A firm should only be publishable if it is valid, has an office and is not missing advisers

munckymagik:
# Summary

This change tightens up the rules that govern when a firm record is exposed to the consumer directory data store (ElasticSearch).

My rationale is this: it's easier to guard against (and reason about) invalid data being exposed to the directory with simple checks here in the model, than it is to guard against (or even imagine) all combinations of missing data or weakly specified queries in the directory.

# Rules implemented

1) A firm should not be discoverable in the directory if it offers face-to-face advice but does not have any advisers. In this state the firm should not be considered publishable.

2) It is possible to add offices and advisers to a firm before the firm record has been completely filled out. In this state the firm should not be considered publishable until it is also valid.

# Notable changes

After implementing the change:

* `Firm#publishable?` now implements the above logic in addition to the existing check for at least one office.
* There is a new `Firm#missing_advisers?` method that provides a home for the logic to check whether the firm is a face-to-face without any advisers.
* The default `:firm` factory now creates 1 adviser by default so we can rely on fixtures being publishable (unless we explicitly want otherwise)
* The `:onboarded_firm` factory is no longer different to the default `:firm` factory, so it becomes an alias of the latter.
* There is a new `:firm_without_advisers` factory for convenience and clarity for those tests that need it (there turned out to be several).

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -199,7 +199,7 @@ class Firm < ActiveRecord::Base
   end
 
   def missing_advisers?
-    primary_advice_method == :local && advisers.empty?
+    (primary_advice_method == :local) && advisers.empty?
   end
 
   private
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -195,7 +195,7 @@ class Firm < ActiveRecord::Base
   end
 
   def publishable?
-    main_office.present?
+    valid? && main_office.present?
   end
 
   private
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -195,7 +195,11 @@ class Firm < ActiveRecord::Base
   end
 
   def publishable?
-    valid? && main_office.present?
+    valid? && main_office.present? && !missing_advisers?
+  end
+
+  def missing_advisers?
+    primary_advice_method == :local && advisers.empty?
   end
 
   private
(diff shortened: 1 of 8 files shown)


JiggyPete:
:+1

urfolomeus:
:+1: whilst it feels like `#missing_advisers?` would read better as `#has_advisers?` (or something that Rubocop will allow), I can see why you've done it this way round.

munckymagik:
@urfolomeus Maybe the name isn't clear yet hmmm.

It's more like `needs_advisers_but_hasnt_got_any?`. Thoughts?

urfolomeus:
`#at_least_one_adviser?` ?

munckymagik:
@urfolomeus I'm trying to bring out the fact that it may not have any advisers and that's ok when the firm is remote-advice-only. But not when it's face-to-face when it definitely needs at_least_one_adviser.

urfolomeus:
`#requires_an_adviser?` ?

munckymagik:
¯\_(ツ)_/¯

munckymagik:
FYI I've run the tests in both rad and rad_consumer against this branch and all pass.

urfolomeus:
totes amazz

darokel:
Just one stylistic comment on there but :+1: otherwise

darokel:
:+1:

munckymagik:
Thanks folks :)

PR #102, opened 2015-09-17
We must use after commit hook for indexing to ElasticSearch

munckymagik:
# The problem

We cannot use the `after_save` callback for reindexing. From the [Rails docs on after_save](http://apidock.com/rails/ActiveRecord/Callbacks/after_save)

`"Note that this callback is still wrapped in the transaction around save. For example, if you invoke an external indexer at this point it won’t see the changes in the database."`

# What has changed?

1. `Firm#publish_to_elastic_search` is now called on `after_commit` again unless the firm is being destroyed.
2. Once the above was done, reindexing the previous parent firm for an adviser being moved stopped working as the `firm_id_changed?` method will return false after the transaction has been committed. So reverted previous commit to fix this behaviour.

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -64,7 +64,8 @@ class Adviser < ActiveRecord::Base
   private
 
   # All record of what changed is gone by the time we get to the after_commit
-  # hooks, so we need to store any important changes here to be actioned later.
+  # hooks. So we cannot use #firm_id_changed? at that point. To work around
+  # this we flag any important changes here to be actioned later.
   def flag_changes_for_after_commit
     @old_firm_id = firm_id_change.first if firm_id_changed?
   end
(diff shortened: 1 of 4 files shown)


app/models/adviser.rb
```diff
@@ -63,6 +63,12 @@ def field_order
 
   private
 
+  # All record of what changed is gone by the time we get to the after_commit
+  # hooks, so we need to store any important changes here to be actioned later.
+  def flag_changes_for_after_commit
```
urfolomeus: Is it worth explicitly calling out that the `#.firm_id_changed?` method will not work? I remember reading a comment similar to this one (maybe even exactly this one) and thinking "oh but surely you can just use ..." (like an idiot). :)

urfolomeus:
:+1:

app/models/adviser.rb
munckymagik: Yes ok.

darokel:
:+1:

munckymagik:
Thanks guys.

PR #103, opened 2015-09-28
Ensure firms are removed from ElasticSearch if they become unpublishable

munckymagik:
Although the exiting code guarded against firms being published before
they are ready, it did not deal with the case where a firm becomes
unpublishable again at some point after being published.

New logic added to delete the entry from ElasticSearch if this happens.

### DIFF
--- a/app/jobs/index_firm_job.rb
+++ b/app/jobs/index_firm_job.rb
@@ -4,6 +4,10 @@ class IndexFirmJob < ActiveJob::Base
   end
 
   def perform(firm)
-    FirmRepository.new.store(firm) if firm.publishable?
+    if firm.publishable?
+      FirmRepository.new.store(firm)
+    else
+      DeleteFirmJob.perform_later(firm.id)
+    end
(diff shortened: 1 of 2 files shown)


app/jobs/index_firm_job.rb
```diff
@@ -4,6 +4,10 @@ class IndexFirmJob < ActiveJob::Base
   end
 
   def perform(firm)
-    FirmRepository.new.store(firm) if firm.publishable?
+    if firm.publishable?
+      FirmRepository.new.store(firm)
+    else
+      DeleteFirmJob.perform_later(firm.id)
```
JiggyPete: My first thought here was- what happens if we try to delete something that doesn't exist in Elastic Search. Digging down into the ElasticSearchClient, it just returns if the delete HTTP request was ok or not. So looks good!

JiggyPete:
:+1:

rnalexander:
:+1: :shipit:

PR #104, opened 2015-09-28
Reimplement onboarding in terms of publishable

munckymagik:
Despite originally thinking that quite fine-grained messaging would be required for the onboarding message in RAD self-service, the reality is there is a single message shown until any one firm becomes publishable. And the only method called by RAD was the boolean `Firm#onboarded?`.

So this change basically strips out all the unused code and tests and reimplements `Firm#onboarded?` to ask each of the principal's firms, "are you publishable?". If any answer "yes" then the onboarded message no longer needs to be shown.

This also resolves the problem that offices were not being taken into account. As `Firm#publishable?` does check for offices, this problem is now fixed too.

If this PR is accepted another will be raised to optimise `Firm#publishable?`, but I want to deal with that separately.


spec/models/principal_spec.rb
```diff
@@ -204,221 +204,35 @@
...
-      end
-    end
-
-    context 'when a principal has not been onboarded' do
-      before :each do
-        attrs = FactoryGirl.attributes_for :not_onboarded_firm, fca_number: principal.fca_number
-        principal.firm.update_attributes(attrs)
+        expect(principal.main_firm_with_trading_names).to have(2).items
+        expect(principal.main_firm_with_trading_names)
+          .to all(have_attributes(publishable?: false))
```
JiggyPete: Didn't know about `have_attributes`. Sweet

JiggyPete:
:+1 Great work, lots of cack cleaned up. Double :+1: for killing the state machine. Don't tell @urfolomeus :-)

rnalexander:
Nice one :+1:

PR #105, opened 2015-09-29
Optimise Firm#publishable?

munckymagik:
When PR https://github.com/moneyadviceservice/mas-rad_core/pull/104 gets merged, `Firm#publishable?` will get called for every firm/trading name associated with a principal, for every page of self-service that should show an onboarding message. Therefore I decided to take this opportunity to optimise it.

The main problem seemed to be calling `#valid?`. By using the `#registered?` method instead, which is basically a heuristic check for validity, `publishable?` now runs in 1/10th of the time.

On benchmarking with a firm that has 100 invalid trading names the total system+user CPU time when calling Principal#onboarded? was about 1.0 second.

After making these changes this went down to 0.1 seconds.

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -195,7 +195,7 @@ class Firm < ActiveRecord::Base
   end
 
   def publishable?
-    valid? && main_office.present? && !missing_advisers?
+    registered? && main_office.present? && !missing_advisers?
   end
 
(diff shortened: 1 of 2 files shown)


benbarnett:
:+1:

rnalexander:
:+1: :shipit:

PR #106, opened 2015-10-01
Use email and phone fields from main_office and remove from firm

munckymagik:
## Summary

There is no longer a good reason to have email and phone contact details in the firm record.

There are too many of these fields in the system (they exist on Principal, Firm and Office) which feels confusing. So instead we now use the details from the `Firm#main_office` and remove the fields from Firm.

## Oh, look how small that ice berg is ...

Unfortunately the `Firm#email_address` field had taken on special extra magical powers in that it was the field used to detect firms that had been fully filled out and saved with validation on (see `Firm#registered?`).

Turns out there aren't many mandatory but simple/scalar Firm fields left that we can use for this trick. Luckily we can use `free_initial_meeting` instead.

## Next steps

Tests will need to be fixed in rad and rad_consumer.

In rad particularly, knowledge has leaked into the tests about the `Firm#email_address` field's magic powers. So I'm going raise another PR after this which encapsulates the `free_initial_meeting` field's new role so we don't get the same leak again.

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -10,7 +10,7 @@ class Firm < ActiveRecord::Base
     :wills_and_probate_flag
   ]
 
-  scope :registered, -> { where.not(email_address: nil) }
+  scope :registered, -> { where.not(free_initial_meeting: nil) }
   scope :sorted_by_registered_name, -> { order(:registered_name) }
 
   has_and_belongs_to_many :in_person_advice_methods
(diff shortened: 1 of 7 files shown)


benbarnett:
:+1:

pmcostello:
:+1:

JiggyPete:
:+1: Well done, made it all look so simple!

PR #107, opened 2015-10-01
[DEV APPROVED] Attempt to encapsulate the registered marker field inside of Firm (wdyt?)

munckymagik:
As alluded to in https://github.com/moneyadviceservice/mas-rad_core/pull/106, this PR is an attempt to stop magic knowledge spreading about the code bases.

In this change, there is now one and only one place that knows which field in Firm is being used to guess if a record has been fully filled out and saved/validated; the Firm object itself.

## wdyt?

Really interested in opinions on whether this is easier to understand or whether there is a better way to do this.

The only thing I'd like to retain is something like the `Firm#__set_registered` method as I've already started crafting a branch of test fixes on RAD which use it. It is only defined when in `RAILS_ENV=test` and seems to hide the magic nicely.

Also, does the double underscore in `__set_registered` successfully communicate, `"beware this is not normal production compatible code you are using, understand clearly please"`? Because that's what I'm trying to indicate.
`¯\_(ツ)_/¯`

UPDATE you can look at https://github.com/moneyadviceservice/rad/pull/318 to see how `__set_registered` is being used.

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -1,6 +1,12 @@
 class Firm < ActiveRecord::Base
   include Geocodable
 
+  FREE_INITIAL_MEETING_VALID_VALUES = [true, false].freeze
+
+  # We use a scalar required field as a marker to detect a record saved with validation
+  REGISTERED_MARKER_FIELD = :free_initial_meeting
+  REGISTERED_MARKER_FIELD_VALID_VALUES = FREE_INITIAL_MEETING_VALID_VALUES
+
   ADVICE_TYPES_ATTRIBUTES = [
     :retirement_income_products_flag,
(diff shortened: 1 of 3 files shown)


pmcostello:
:+1:

rnalexander:
:+1: :shipit: 

It looks ok.  Not sure about the double underscores but it's better than naming it normally.  I know it's done in Javascript but the jury seems to still be out about that as a convention in Ruby.  But it's a good attempt to mitigate confusion and hopefully bump up discoverability.

mconnell:
:+1:

munckymagik:
Thanks @rnalexander. Just had a similar conversation here with @mconnell.

PR #109, opened 2015-10-14
[DEV APPROVED] add system_name method similiar to friendly_name

mconnell:
This PR introduces the notion of a 'system' name.

Currently, things like the Advice methods for an adviser ("Phone", "Online") are stored as records in the database (`OtherAdviceMethod` objects). These objects have a defined order value associated with them. The significance of this order value has importance when trying to fetch I18n translations for the associated records.

For example:

In the current implementation of fetching text for one of these records, there is a yaml key like `'en.other_advice_methods.ordinal.1'`, which maps to `'Phone'`. An existing `friendly_name` method makes use of this to allow translations to be rendered. This is ok in isolation, but in scenarios where we need to iterate over a collection of objects, and populate more complex UI design, we're finding we're running up against this implementation.

This change introduces a method `system_name`, which instead of returning a translation, returns a sensibly named symbol that represents this object, that way we have more control over translations and conditional logic within any application which uses rad_core.

This also has an advantage of us having these known order numbers as constants on the associated class, instead of being hidden away in translations file.

## how it's used
In the scenario we have an `OtherAdviceMethod` object with an `id` of `5`, but an `order` value of '1'
```ruby
class OtherAdviceMethod
  include SystemNameable
  SYSTEM_NAMES = { 1 => :phone, 2 => :online }
end
```

`OtherAdviceMethod.system_name(5)`, will return `:phone`.

We can then use this symbol in specific translations, rendering of specific assets, or setting up conditionals where needed.

### DIFF
--- a/app/models/other_advice_method.rb
+++ b/app/models/other_advice_method.rb
@@ -1,6 +1,12 @@
 class OtherAdviceMethod < ActiveRecord::Base
   include Translatable
   include FriendlyNamable
+  include SystemNameable
+
+  SYSTEM_NAMES = {
+    1 => :phone,
+    2 => :online
+  }
 
   has_and_belongs_to_many :firms
 
--- a/app/models/system_nameable.rb
+++ b/app/models/system_nameable.rb
@@ -1,0 +1,12 @@
+module SystemNameable
+  def self.included(base)
+    base.extend(ClassMethods)
+  end
+
+  module ClassMethods
+    def system_name(id)
+      order = find(id).order
+      self::SYSTEM_NAMES.fetch(order, :not_found)
+    end
+  end
+end
--- a/spec/models/other_advice_method_spec.rb
+++ b/spec/models/other_advice_method_spec.rb
@@ -2,4 +2,5 @@ RSpec.describe OtherAdviceMethod do
   it_behaves_like 'reference data'
   it_behaves_like 'translatable'
   it_behaves_like 'friendly named'
+  it_behaves_like 'system named'
 end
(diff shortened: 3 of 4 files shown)


munckymagik:
:+1:

spencer:
:+1:

darokel:
:+1:

munckymagik:
Thank you folks!

asmega:
why not use and instance method? e.g. `some_advice_method.system_name` => `:phone`.

i just think i'm missing some context on how this is used

munckymagik:
@asmega yes Mark and I discussed adding this, but we just don't have an actual use case for it just now. Easy enough to add when we need it. For context we have WIP PR using it here: https://github.com/moneyadviceservice/rad_consumer/pull/220/files#diff-10b08810a5f18dc3ea24f8e63a4a3de8R38

asmega:
personally i would make that array return an object on each iteration instead of an int. you have to do a look up each time anyway for the system name. then you can call `advice_method.system_name`. but that's up to you if you prefer as a class method.

it will also look more obvious that you need to eager load the association

munckymagik:
@asmega I get what you're saying. It's not idea but we're working in a context where all data has come via ElasticSearch rather than the DB. This decoupling has been very useful in the past so we're sticking with it. Consequently there will be slightly inverted code like this occasionally.

`firm.other_advice_methods` is a list of ids which comes from ElasticSearch.

asmega:
ok i see, i just saw that record was an ActiveRecord so assumed it was from the database

munckymagik:
@asmega Yeah, it's a bit confusing without context. I keep having to remind myself when I see `firm` in rad it's an `ActiveRecord`, but when I see it in rad_consumer it's a `FirmResult` from ElasticSearch.

PR #110, opened 2015-10-14
[DEV APPROVED] Update Qualifications/Accreditation factories to generate orders from 1

munckymagik:
This was the only change needed to implement the Qualification/Accreditation filter in rad_consumer

The seeds file in RAD generates the order using .each.with_index(1) so
the indexes generated start at 1.

https://github.com/moneyadviceservice/rad/blob/master/db/seeds.rb#L7
https://github.com/moneyadviceservice/rad/blob/master/db/seeds.rb#L73

So needed to update the factories to do the same.

Checked the tests in rad against this change and found no issues.

### DIFF
--- a/spec/factories/accreditation.rb
+++ b/spec/factories/accreditation.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :accreditation do
     sequence(:name) { |n| "Accreditation #{n}" }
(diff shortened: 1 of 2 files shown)


mconnell:
:+1:

benbarnett:
:+1:

PR #112, opened 2015-10-20
add a method to allow us to check types of advice

mconnell:
Instead of relying on a magic number in rad_consumer like so:

```ruby
@firm_result.retirement_income_products == 100
```
We replace it with a method that allows us to ask any of the advice types if they have been included in the firm's results.

```ruby
@firm_result.includes_advice_type? :retirement_income_products
```
This is in response to a request in a PR for rad consumer so we aren't littering a magic number around the codebases: https://github.com/moneyadviceservice/rad_consumer/pull/225#discussion_r42373929

### DIFF
--- a/lib/mas/firm_result.rb
+++ b/lib/mas/firm_result.rb
@@ -1,6 +1,8 @@
 require 'uk_phone_numbers'
 
 class FirmResult
+  PERCENTAGE_FOR_TRUE = 100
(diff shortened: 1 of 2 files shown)


benbarnett:
:+1:

JiggyPete:
:+1: nice

PR #113, opened 2015-10-20
[DEV APPROVED] Separate the geocoding from the geocoding background jobs + name clarification

munckymagik:
## Motivation

This is the first in a series of PRs intended to make the Office models geocodable. The first few will be refactoring to make that goal easier.

1. Geocoding is relatively inexpensive despite requiring a 3rd party webservice. Why? Because every postcode based search on the consumer frontend makes an inline/blocking request to geocode the postcode given. If there was a problem with that we would soonest see it there.
2. Geocoding models on a background job means geocoding failures don't get reported anywhere, so the user has no idea there's a problem.

This PR separates out the act of geocoding from the GeocodeXJob classes so we have the opportunity to geocode in-request. But also retains the existing background jobs (for now).

## Second issue

`Geocodable#geocode!` feels more like it should be called `update_coordinates!`. So I renamed it as such. Totally open to suggestion for a better name.

### DIFF
--- a/spec/lib/mas/model_geocoder_spec.rb
+++ b/spec/lib/mas/model_geocoder_spec.rb
@@ -1,0 +1,81 @@
+RSpec.describe ModelGeocoder do
+  let(:model_class) do
+    Class.new do
+      attr_accessor :address_line_one, :address_line_two, :address_postcode, :longitude, :latitude
+
+      def geocode!(*args); end
+
+      def full_street_address
+        [address_line_one, address_line_two, address_postcode, 'United Kingdom'].reject(&:blank?).join(', ')
+      end
+    end
+  end
+
+  let(:model) do
+    model_class.new.tap do |thing|
+      thing.address_line_one = address_line_one
+      thing.address_line_two = address_line_two
+      thing.address_postcode = address_postcode
+    end
+  end
+
+  let(:address_line_one) { '120 Holborn' }
+  let(:address_line_two) { 'London' }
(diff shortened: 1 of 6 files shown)


spec/lib/mas/model_geocoder_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  let(:address_line_one) { '120 Holborn' }
+  let(:address_line_two) { 'London' }
+  let(:address_postcode) { 'EC1N 2TD' }
+  let(:expected_coordinates) { [51.5180697, -0.1085203] }
+
+  describe '#geocode' do
+    context 'when the model address can be geocoded' do
+      it 'returns the coordinates' do
+        VCR.use_cassette('geocode-one-result') do
```
JiggyPete: Much prefer this style for using VCR cassettes!

munckymagik: Thanks @JiggyPete. Do you like it enough to give this thumb though? ;)

tomas-stefano:
:+1: :+1:

spec/lib/mas/model_geocoder_spec.rb
JiggyPete: I do indeed. Held off yesterday to chat about `:geocode!` then promptly forgot 😀

JiggyPete:
:+1:

darokel:
:+1:

munckymagik:
Thanks people!

PR #114, opened 2015-10-21
[DEV APPROVED] Clarify the rules to interpret the advice type numbers

munckymagik:
These numbers used to be percentage values but they were converted to booleans. However, a wide range of number values still exist in the production index, so strictly we need to interpret 0 as not selected and > 0 as selected. This intention can be confirmed by looking at the migration when the work was originally done:

https://github.com/moneyadviceservice/mas-rad_core/blob/master/db/migrate/20150630143001_change_percentage_fields_to_boolean_fields_on_firm.rb#L16

### DIFF
--- a/lib/mas/firm_result.rb
+++ b/lib/mas/firm_result.rb
@@ -1,7 +1,7 @@
 require 'uk_phone_numbers'
 
 class FirmResult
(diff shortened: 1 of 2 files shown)


darokel:
:+1:

urfolomeus:
:+1:

PR #115, opened 2015-10-22
We shouldn't be calling translations across code bases

urfolomeus:
We needed to change the translation for the new design but then realised that it was actually being called from within the gem. The data should not be using the translations (or formatting for that matter) here and so we've moved it back to rad_consumer. The raw value is still passed through from here though.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    mas-rad_core (0.0.85)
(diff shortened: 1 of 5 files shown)


munckymagik:
:+1: once the Travis shenanigans have been de-shenanignized.

benbarnett:
:+1:

urfolomeus:
Thanks @munckymagik @benbarnett

PR #117, opened 2015-11-02
Update firm minimum fixed fee defaults to zero

JiggyPete:
Where no minimum fee has been explicitly set by the firms the field remains unset and the directory does not show any information on fees. So for a long time the RAD admin team have been manually updating records to put in £0 whenever they have the opportunity.

Rather than them having to do drive-by updates forever, this change solves this in one shot by setting a default value.

### DIFF
--- a/db/migrate/20151102113518_add_default_value_to_minimum_fixed_fee.rb
+++ b/db/migrate/20151102113518_add_default_value_to_minimum_fixed_fee.rb
@@ -1,0 +1,15 @@
+class AddDefaultValueToMinimumFixedFee < ActiveRecord::Migration
+  class Firm < ActiveRecord::Base
+  end
+
+  def up
+    firms = Firm.where(minimum_fixed_fee: nil)
+    firms.each { |f| f.update_attribute('minimum_fixed_fee', 0) }
+
+    change_column_default :firms, :minimum_fixed_fee, 0
+  end
+
+  def down
+    change_column_default :firms, :minimum_fixed_fee, nil
+  end
+end
--- a/spec/models/firm_spec.rb
+++ b/spec/models/firm_spec.rb
@@ -324,10 +324,38 @@ RSpec.describe Firm do
     end
 
     describe 'minimum fixed fee' do
+      context 'default value' do
+        it { expect(Firm.new.minimum_fixed_fee).to eq(0) }
+      end
+
       context 'when not numeric' do
         before { firm.minimum_fixed_fee = 'not-numeric' }
 
         it { is_expected.not_to be_valid }
+      end
+
+      context 'when blank' do
+        before { firm.minimum_fixed_fee = nil }
+
+        it { is_expected.not_to be_valid }
+      end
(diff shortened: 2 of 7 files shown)


db/migrate/20151102113518_add_default_value_to_minimum_fixed_fee.rb
```diff
@@ -0,0 +1,12 @@
+class AddDefaultValueToMinimumFixedFee < ActiveRecord::Migration
+  def up
+    firms = Firm.where(minimum_fixed_fee: nil)
+    firms.each{|f| f.update_attribute('minimum_fixed_fee', 0)}
```
munckymagik: I know we don't have Rubocop on this repo yet, but formatting here needs spaces `each { |f| .... 0) }`

JiggyPete: Rubocop is now happy with this line.

db/migrate/20151102113518_add_default_value_to_minimum_fixed_fee.rb
```diff
@@ -0,0 +1,12 @@
+class AddDefaultValueToMinimumFixedFee < ActiveRecord::Migration
+  def up
+    firms = Firm.where(minimum_fixed_fee: nil)
```
urfolomeus: You're using Firm here but it might not exist. We should add `class Firm < ActiveRecord::Base; end` on line 2.

spec/models/firm_spec.rb
```diff
@@ -324,11 +324,33 @@ def set_marker_field(firm, value)
...
+        before { firm.minimum_fixed_fee = 0 }
+
+        it { is_expected.to be_valid }
+      end
+
+      context 'when more than zero' do
+        before { firm.minimum_fixed_fee = 1 }
+
+        it { is_expected.to be_valid }
+      end
```
urfolomeus: What about if it's less than zero? Should this be valid or invalid?

db/migrate/20151102113518_add_default_value_to_minimum_fixed_fee.rb
JiggyPete: Will that need an `attr_accessible :minimum_fixed_fee` incase the column doesn't exist?

urfolomeus: No, we only care if it's in the database or not. It's just that we can't access the database if we don't have an AR model for it :)

spec/models/firm_spec.rb
JiggyPete: @urfolomeus all updated, can you take another look please?

urfolomeus:
:+1:

JiggyPete:
@munckymagik cleaned up the schema.rb

munckymagik:
:+1:

asmega:
:+1:

PR #119, opened 2015-11-05
[DEV APPROVED] Geocoding offices

munckymagik:
## Summary

The PR introduces changes to the office model to make it geocodable.

As previously mentioned in #113, geocoding is quick enough that it can be done within the request lifecycle rather than a background job. This has the added benefit that geocoding failures can be reported to the user (which for advisers they currently aren't as they are done on a background job).

So the design here provides a way to geocode synchronously.

## API Examples

Geocoding an in-memory office instance is done by calling:

```ruby
office.geocode
office.changed? # => true, record is dirty because lat/long are set but the record is not saved at this point
``` 

Geocoding and saving can be done by:

```ruby
office = firm.offices.build(office_params)
office.save_with_geocoding # => true if geocoded and saved successfully
```

Or updating as:

```ruby
office.update_with_geocoding(office_params) # => true if geocoded and saved successfully
```

Failure to geocode invalidates the record just like normal validations would so we can report in the UI using existing validations infrastructure:

```ruby
office.update_with_geocoding({ address_postcode: 'XX1 1XX' }) # => false, geocoding fails
office.errors.include?(:address) # => true
```

## Rationale

One thing I wrestled with was whether to try to peg geocoding onto one of ActiveRecord's standard callbacks. But I couldn't settle on a callback to use that didn't have some kind of draw back.

One big issue I had with this was the amount of "surprise" using callbacks causes. E.g. a maintainer might be surprised to find that `office.save` would have the side-effect of issuing an HTTP request to a remote service, while calling `office.save(validate: false)` wouldn't.

Therefore I chose to go with an explicit interface. If you want to geocode then you call `save_with_geocoding` or `update_with_geocoding` which both do what they claim. Both require validation to happen and don't provide a way to let calling code opt-out like `save` does.

## Outstanding Issues

* At this point I've disabled the code which triggers reindexing when changes to an office occur. This PR would be a lot bigger and harder to understand with that included. I plan to get this work approved but not merge it until I have a PR for indexing too.
* I have approval from the product team to retrofit this geocoding strategy to the `Adviser` records so a lot of this new code will be refactored out of `Office` so it can be shared (probably into `Geocodable`).

### DIFF
--- a/app/models/geocodable.rb
+++ b/app/models/geocodable.rb
@@ -21,8 +21,12 @@ module Geocodable
     [latitude, longitude]
   end
 
-  def update_coordinates!(coordinates)
+  def coordinates=(coordinates)
     self.latitude, self.longitude = coordinates
+  end
+
+  def update_coordinates!(coordinates)
+    self.coordinates = coordinates
     update_columns(latitude: latitude, longitude: longitude)
   end
 end
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -95,7 +95,7 @@ class Office < ActiveRecord::Base
   private
 
   def add_geocoding_failed_error
-    errors.add(:address, 'could not be geocoded')
+    errors.add(:address, I18n.t("#{model_name.i18n_key}.geocoding.failure_message"))
   end
 
   def upcase_postcode
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -1,6 +1,14 @@
 class Office < ActiveRecord::Base
   include Geocodable
 
+  ADDRESS_FIELDS = [
+    :address_line_one,
+    :address_line_two,
+    :address_town,
+    :address_county,
+    :address_postcode
+  ].freeze
+
   belongs_to :firm
 
   before_validation :upcase_postcode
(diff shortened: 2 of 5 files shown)


app/models/geocodable.rb
```diff
@@ -21,8 +21,12 @@ def coordinates
...
 
-  def update_coordinates!(coordinates)
+  def coordinates=(coordinates)
     self.latitude, self.longitude = coordinates
-    update_columns(latitude: latitude, longitude: longitude)
+  end
+
+  def update_coordinates!(coordinates)
+    self.coordinates = coordinates
+    update_columns(latitude: latitude, longitude: longitude) if persisted?
```
mconnell: I don't think you need the `if persisted?`

munckymagik: Thanks as discussed. `if persisted?` should NOT be here, the need for it was replaced by `def coordinates=` but I clearly forgot to go back and take it out :( my bad.

mconnell:
Other than the thing I pointed out, that you then pointed out, that I then made a comment about... :+1:

app/models/office.rb
```diff
@@ -57,21 +65,41 @@ def full_street_address
...
+
+  def update_with_geocoding(office_params)
+    self.attributes = office_params
+    save_with_geocoding
+  end
+
+  private
+
+  def add_geocoding_failed_error
+    errors.add(:address, 'could not be geocoded')
```
munckymagik: Also note this hard coded error message will be replaced by a translation in an upcoming PR.

rnalexander:
:+1:

munckymagik:
I'm going to use this branch as a temporary 'master-for-geocoding-changes' so will close this rather than merge at this point. Will raise a final PR to merge all the approved parts to master when that moment arrives.

PR #120, opened 2015-11-05
[DEV APPROVED] Factor out shareable methods to GeocodableSync (temporary name)

munckymagik:
## Summary

This PR builds upon #119 to move the code that relates to geocoding out of Office to a mixin that will be shareable with Adviser.

## Rationale

At this point it makes the office_spec file a bit easier to manage again and paves the way for what is to come.

## Future changes

* Once the Adviser class uses this mixin too, we can merge GeocodableSync into Geocodable and nuke Geocodable#update_coordinates! (as it does horrible things like saving changes without validation, bypasses the dirty state methods and requires a persisted object or it won't work)
* Combine the tests for both


rnalexander:
:+1:

mconnell:
:+1:

PR #121, opened 2015-11-09
[DEV APPROVED] Firm indexer

munckymagik:
## Summary

This PR introduces a home for the logic relating to indexing firms. In the existing system this was spread around between advisers, firms and the geocoding jobs and the index firm job.

The `FirmIndexer` (for want of a better name) is now *THE* only thing that knows the logic about when and how a firm is indexed to ElasticSearch. The models simplify notify it in `on_commit` handlers that something has changed and it then works out the required outcome from what it gets given.

## Consequences

At this point indexing is no longer done on an ActiveJob. We could easily put this back in (which is why I've not deleted the job classes in this PR), but we need to prove that doing this on a background job is actually necessary.

## Next work

* The `Adviser` has not yet been fully converted over. That will be done next.
* The `Firm` still has lat/long fields, includes Geocodable, and delegates address fields to main office. This needs to be taken out.
* I'd feel more comfortable if we had a integration test that exercises a few of the indexing/geocoding scenarios end to end.
* We need to expose the office coordinates to ElasticSearch.
* We need to think about migrating existing geo data from firms to the main offices.

### DIFF
--- a/lib/mas/firm_indexer.rb
+++ b/lib/mas/firm_indexer.rb
@@ -16,17 +16,19 @@ module FirmIndexer
       index_firm(aggregate.firm) if firm_exists?(aggregate.firm)
     end
 
+    def firm_exists?(firm)
+      return false if (firm.nil? || firm.destroyed?)
+      Firm.exists?(firm.id)
+    end
+
+    private
+
     def store_firm(firm)
       FirmRepository.new.store(firm)
     end
 
     def delete_firm(firm)
       FirmRepository.new.delete(firm.id)
-    end
-
-    def firm_exists?(firm)
-      return false if (firm.nil? || firm.destroyed?)
-      Firm.exists?(firm.id)
     end
   end
 end
--- a/lib/mas/firm_indexer.rb
+++ b/lib/mas/firm_indexer.rb
@@ -1,14 +1,22 @@
 module FirmIndexer
   def self.index_firm(firm)
-    repo = FirmRepository.new
-
     if firm.publishable?
-      repo.store(firm)
+      store_firm(firm)
     else
-      repo.delete(firm.id)
+      delete_firm(firm)
     end
   end
 
   def self.handle_firm_changed(firm)
+    return delete_firm(firm) if firm.destroyed?
+    index_firm(firm)
+  end
+
+  def self.store_firm(firm)
+    FirmRepository.new.store(firm)
+  end
+
+  def self.delete_firm(firm)
+    FirmRepository.new.delete(firm.id)
   end
 end
--- a/lib/mas/firm_indexer.rb
+++ b/lib/mas/firm_indexer.rb
@@ -11,12 +11,17 @@ module FirmIndexer
     alias_method :handle_firm_changed, :index_firm
 
     def handle_aggregate_changed(aggregate)
-      index_firm(aggregate.firm) if firm_exists?(aggregate.firm)
+      # This method may be invoked as part of a cascade delete, in which case
+      # we should do nothing here. The firm change notification will handle
+      # the change.
+      return if associated_firm_destroyed?(aggregate)
+      index_firm(aggregate.firm)
     end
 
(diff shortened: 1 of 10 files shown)


lib/mas/firm_indexer.rb
```diff
@@ -0,0 +1,32 @@
...
+        delete_firm(firm)
+      end
+    end
+
+    def handle_firm_changed(firm)
+      index_firm(firm)
+    end
+
+    def handle_aggregate_changed(aggregate)
+      index_firm(aggregate.firm) if firm_exists?(aggregate.firm)
```
urfolomeus: Would it be beneficial to make this conditional a guard in `index_firm` instead? It is currently possible to pass something that is not an existing Firm to that method.

lib/mas/firm_indexer.rb
```diff
@@ -0,0 +1,32 @@
...
+  class << self
+    def index_firm(firm)
+      if !firm.destroyed? && firm.publishable?
+        store_firm(firm)
+      else
+        delete_firm(firm)
+      end
+    end
+
+    def handle_firm_changed(firm)
```
urfolomeus: Is this not just an alias?

lib/mas/firm_indexer.rb
```diff
@@ -0,0 +1,32 @@
...
+      index_firm(firm)
+    end
+
+    def handle_aggregate_changed(aggregate)
+      index_firm(aggregate.firm) if firm_exists?(aggregate.firm)
+    end
+
+    def store_firm(firm)
```
urfolomeus: Should the methods from this point down perhaps be private?

lib/mas/firm_indexer.rb
munckymagik: This is very clear is it. But there's a difference in context.

In `index_firm` if the firm is "destroyed" we want to remove it from the index.

In `handle_aggregate_changed` if the firm is destroyed/doesn't exist we want to do nothing.

If we move the guard into `index_firm` we would need to work out whether it was coming from `handle_aggregate_changed`.

Maybe `firm_exists?` is a poor name. What it is being used to express is, "are we in the middle of a cascade delete?"

lib/mas/firm_indexer.rb
munckymagik: Yes I'll have a shot a making it an alias.

lib/mas/firm_indexer.rb
munckymagik: I see what you mean. They remained public to this point because I wanted to be able to test `firm_exists?` so I could simplify the test for `handle_aggregate_changed?` and mock store/delete_firm. I'll have a rethink at this point and see.

urfolomeus:
:+1:

JiggyPete:
:+1:

spec/lib/mas/firm_indexer_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    let(:aggregate) { firm.offices.first }
+    subject { described_class.associated_firm_destroyed?(aggregate) }
+
+    context 'when the firm instance and db record exist' do
+      it { is_expected.to be(false) }
+    end
+
+    context 'when firm has been destroyed before the association has been loaded' do
+      before do
+        aggregate # load the aggregate, but not the aggregate.firm association
```
asmega: does `let!(:aggregate)` fix this as this removes the laziness

asmega:
:+1:

spec/lib/mas/firm_indexer_spec.rb
munckymagik: Yes I deliberated about using that as a solution, but I wanted to communicate very clearly what was going on and felt this way did that. wdyt?

asmega: yeah im find with that

munckymagik:
Thanks @asmega

PR #123, opened 2015-11-10
Add investing booleans to firm result

mconnell:
This change enables the Sharia and Ethical investing flags to be serialised for elastic search, and also available in a `FirmResult`, which is the object that represents deserialised data from elastic search.


munckymagik:
:+1:

rnalexander:
:+1: :shipit:

PR #127, opened 2015-11-10
[DEV APPROVED] Geocode offices on postcode alone

munckymagik:
After discussing with Phil and Seb a decision was made to geocode offices on postcode alone. The idea being to lower the chances of geocoding failures by reducing the number of variables that affect it.

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -2,14 +2,6 @@ class Office < ActiveRecord::Base
   include Geocodable
   include GeocodableSync
 
-  ADDRESS_FIELDS = [
-    :address_line_one,
-    :address_line_two,
-    :address_town,
-    :address_county,
-    :address_postcode
-  ].freeze
-
   belongs_to :firm
 
   before_validation :upcase_postcode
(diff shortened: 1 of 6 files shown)


app/models/office.rb
```diff
@@ -69,11 +61,11 @@ def telephone_number
   end
 
   def full_street_address
-    [address_line_one, address_line_two, address_postcode, 'United Kingdom'].reject(&:blank?).join(', ')
+    "#{address_postcode}, United Kingdom"
   end
 
   def has_address_changes?
-    ADDRESS_FIELDS.any? { |field| changed_attributes.include? field }
+    changed_attributes.include? :address_postcode
```
asmega: fyi `address_postcode_changed?` should work here

asmega:
:+1:

app/models/office.rb
munckymagik: Nice ... I thought. Then I tried to make this change, nothing's ever simple.

So the way it is right now, it returns either `true` or `false`. Simple.

But using the attribute method version you get `true`, `false` or `nil`.

```
[9] pry(#<Office>)> [changed_attributes.include?(:address_postcode), address_postcode_changed?]
=> [false, nil]
```

I've just spent 5 minutes trying to work out why the `nil` and whether it's me that did something wrong and if that is really the same as `false`. But looking through the source and knowing that `ActiveRecord` overrides bits of `ActiveModel::Dirty` I just decided to stop.

Do you mind if we keep it as is? I could do:

```
!!address_postcode_changed?
```

I feel like I'm being pernickety here, but I just find it easier to think about all this stuff if I can rely on things returning predictable values.

mconnell:
:+1:

PR #134, opened 2015-11-13
Go back to using whole address to geocode offices again

munckymagik:
Reverts most of #127.

The problem we're solving here is this: if firm offices and advisers have the same postcode then the map pins for offices and advisers get plotted on top of each other.

What we found yesterday is that by geocoding offices on the full address, this introduces enough *fuzz* that most offices end up being plotted in a different place, which is enough to work around the problem acceptably. E.g.

pasted image at 2015_11_12 04_48 pm

Now becomes:

pasted image at 2015_11_12 04_47 pm

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -2,6 +2,14 @@ class Office < ActiveRecord::Base
   include Geocodable
   include GeocodableSync
 
+  ADDRESS_FIELDS = [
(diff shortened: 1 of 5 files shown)


Tr4pSt3R:
:+1:

rnalexander:
:+1:

PR #138, opened 2016-02-05
Bypass reference number check

cazrin:
> In some cases, we need to include advisers in the directory who do not have an adviser reference number for us to cross-check against the FCA Register.
>
> This is needed to be able to include equity release advisers who do not have a FCA adviser reference number.
>
> This does not need to be represented in self-service only manual entry by MAS.  This gives us the security that only qualified individuals are being added to directory.

Opening a WIP PR for this piece of work for visibility + comments.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    mas-rad_core (0.0.95)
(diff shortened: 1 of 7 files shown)


munckymagik:
:+1:

Tr4pSt3R:
:+1: very excellent specs

PR #140, opened 2016-02-10
Modify FirmSerializer to serialize Firm#languages

munckymagik:
# What?

Modification to serialize the `Firm#languages` array.

# Why?

This makes firms' languages available in ElasticSearch and therefore means we can surface this data on the RAD directory.

# Background

In the last phase of the RAD we added the ability for firms to specify any non-English languages they can provide their service using. This change is in preparation for showing this information on the consumer facing directory.


tomas-stefano:
LG :+1: :+1:

cazrin:
:+1:

PR #141, opened 2016-02-11
Get languages back out of ElasticSearch

munckymagik:
# What?

Adds the `languages` attribute on the `FirmResult` so that the array of languages are read out of the JSON returned by ElasticSearch.

# Why?

Now that a [firm's spoken languages are serialised to ElasticSearch](https://github.com/moneyadviceservice/mas-rad_core/pull/140) when they save their record, we need a way to get them back out so we can show them in `rad_consumer`.


KateWilkinson:
:+1:

tomas-stefano:
:+1:

munckymagik:
Thanks people.

PR #142, opened 2016-02-15
Report automation: snapshot model to be run daily

cazrin:
Related PR: https://github.com/moneyadviceservice/rad/pull/338

Implementing a new model called `Snapshot`. Intended use is to run a job each day that creates a new record which will run all of the queries on the model. So each Snapshot will store the requested metrics for each day, which we can then query later/export to CSV etc.


cazrin:
Have some errors coming from code climate about a complex class. Going to see what I can do to reduce complexity.

app/models/snapshot.rb
```diff
@@ -0,0 +1,286 @@
...
+    else
+      {}
+    end
+  end
+
+  def publishable_firms
+    @_publishable_firms ||= Firm.registered.select(&:publishable?)
+  end
+
+  def run_queries
```
mconnell: Maybe put a comment beside this to explain the intent (I understand because I've just chatted to you about it)

.codeclimate.yml
```diff
@@ -0,0 +1,2 @@
+exclude_paths:
+- app/models/snapshot.rb
```
tomas-stefano: Is there a reason to exclude this file? The commit message for this change could explain that.

cazrin: I've amended the commit message which hopefully explains my reasoning.

app/models/snapshot/adviser_queries.rb
```diff
@@ -0,0 +1,132 @@
...
+  def query_advisers_in_northern_ireland
+    advisers_in_country(Adviser.all, 'Northern Ireland')
+  end
+
+  TravelDistance.all.keys.each do |val|
+    method_name = val.downcase.gsub(' ', '_')
+    define_method "query_advisers_who_travel_#{method_name}" do
+      advisers_who_travel(val)
+    end
+  end
```
munckymagik: What would happen if we were asked to change the TravelDistances? Would the snapshots cope or would they need to be updated?

munckymagik: Because this makes it look like they would cope with the change.

cazrin: The only thing you would need to do is add a migration to include the new column for the travel distance. `advisers_who_travel_1000_miles` for example.

app/models/snapshot/postcode_finder.rb
```diff
@@ -0,0 +1,36 @@
+require 'net/http'
+
+class Snapshot::PostcodeFinder
+  def country_postcodes(postcodes, country)
```
munckymagik: It's not immediately obvious to me what gets returned by this. Could we change the name in some way to make it clearer?

Also, if PostcodeFinder is not tied to the snapshots, is it something that could go in lib to encourage reuse if we ever need to?

db/migrate/20160217151659_add_missing_travel_distance_to_snapshot.rb
```diff
@@ -0,0 +1,5 @@
+class AddMissingTravelDistanceToSnapshot < ActiveRecord::Migration
```
munckymagik: As (I assume) you are the only one who has run these migrations so far, can we merge these three migrations into one before it hits master?

munckymagik:
:+1:

tomas-stefano:
:+1:

PR #143, opened 2016-02-18
Fixes the query methods not being executed

cazrin:
After moving the query methods to modules, `public_methods(false)` was no longer returning those methods. This snuck in because the specs stub these methods out, and I made the mistake of not checking it in the browser.


mconnell:
:monkey_face: :+1:

rnalexander:
:+1: :shipit: Not sure why why this isn't showing up in reminders...

PR #144, opened 2016-02-19
Make postcode validation more robust

urfolomeus:
There were two issues that we were dealing with here:

1. Entering a postcode without a space between parts fails as invalid.
2. Postcodes with a valid shape but invalid code where being accepted (i.e. 1234 567).

This PR introduces a better postcode validator. It also allows users to add a valid postcode without spaces and will save it with a space in the right place.

demo


app/models/office.rb
```diff
@@ -1,3 +1,5 @@
+require 'uk_postcode'
```
munckymagik: Do we need to require it explicitly?

urfolomeus: Yes, this is not Rails :)

munckymagik: It's not Rails?

urfolomeus: AFAIK there is no way to auto-require from a gemspec?

urfolomeus: It doesn't work if you don't require the gem, so unless there's a way to auto-require that I don't know about ...

spec/support/shared_examples/geocodable_examples.rb
```diff
@@ -14,6 +14,7 @@
 
   def modify_address(subject)
     subject.send("#{address_field_name}=", address_field_updated_value)
+    expect(subject).to have_address_changes
```
munckymagik: Let's change this to `have_changed` so we're using the Rails Dirty API directly. `has_address_changes?` could be broken in the model we're testing.

urfolomeus: Good shout.

app/models/office.rb
munckymagik: Excuse my simplistic approach. What happens if you comment out that line?

urfolomeus: It breaks :)

munckymagik: Fair enough :)

munckymagik:
:+1:

Tr4pSt3R:
:+1:

urfolomeus:
Thanks @munckymagik @Tr4pSt3R :)

PR #145, opened 2016-02-22
[DEV APPROVED] Remove "Other" accreditation: data migration

cazrin:
Related PRs: https://github.com/moneyadviceservice/rad/pull/342 and https://github.com/moneyadviceservice/rad_consumer/pull/271

Was originally in https://github.com/moneyadviceservice/rad/pull/342 but as @munckymagik pointed out it should probably live in `mas-rad_core`.

### DIFF
--- a/db/migrate/20160222091312_remove_other_accreditation.rb
+++ b/db/migrate/20160222091312_remove_other_accreditation.rb
@@ -7,6 +7,13 @@ class RemoveOtherAccreditation < ActiveRecord::Migration
   end
 
   def down
-    fail ActiveRecord::IrreversibleMigration
+    # Unfortunately we can't just insert the 'Other' record, as there are parts
+    # of `rad` and `rad_consumer` which rely on the 'Other' record having the
+    # ID of '5'. This is due to translations and configuring whether or not to
+    # show the record.
+    #
+    # You're allowed to roll back this migration, but be aware that if you're
+    # using an older version of `rad` or `rad_consumer` they may throw errors
+    # if this record doesn't exist.
   end
 end
--- a/db/migrate/20160222091312_remove_other_accreditation.rb
+++ b/db/migrate/20160222091312_remove_other_accreditation.rb
@@ -1,0 +1,12 @@
+class RemoveOtherAccreditation < ActiveRecord::Migration
+  class Accreditation < ActiveRecord::Base
+  end
+
+  def up
+    Accreditation.where(name: 'Other').destroy_all
+  end
+
+  def down
+    fail ActiveRecord::IrreversibleMigration
+  end
+end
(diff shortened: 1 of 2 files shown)


urfolomeus:
:+1:

db/migrate/20160222091312_remove_other_accreditation.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveOtherAccreditation < ActiveRecord::Migration
+  class Accreditation < ActiveRecord::Base
+  end
+
+  def up
+    Accreditation.where(name: 'Other').destroy_all
+  end
+
+  def down
+    fail ActiveRecord::IrreversibleMigration
```
munckymagik: Good idea.

munckymagik:
:+1:

db/migrate/20160222091312_remove_other_accreditation.rb
urfolomeus: Is it really not possible to rollback to a state prior to this migration? Surely it just has no down state?

urfolomeus: Or even the downstate adds back in the Other accreditations from the seed?

cazrin: Possibly. The problem with adding it back in is that all of the other records related to it are hardcoded to the ID ('5'). So adding it back in would probably make it have an ID of 6, and nothing would work anyway.

Is it possible not to have a `down` state? I'm happy to leave that off but not sure of the consequences.

urfolomeus: Yeah you can leave the down state empty or, for clarity, with a comment explaining what you just said :)

munckymagik: It'll be much simpler if we decide with confidence that we're not going to support going backwards.

urfolomeus: The thing is that `fail ActiveRecord::IrreversibleMigration` actually throws an exception and stops the rollback continuing. A comment will explain why you're not putting it back in but let the rollback continue. If the app will break if you rollback passed this migration then the exception is the right thing, otherwise a comment is preferable (IMHO).

cazrin: Happy to go with either, anyone want to make a call? There's already another irreversible migration in this codebase which is why I went with it.

urfolomeus: "If the app will break if you rollback passed this migration then the exception is the right thing, otherwise a comment is preferable (IMHO)." ;)

urfolomeus:
:+1:

rnalexander:
:+1:

PR #146, opened 2016-02-22
[DEV APPROVED]Fix telephone validation: changes to the model 

munckymagik:
# What?

Tightens the validation rules for new telephone numbers on the Office model and centralises the clean up and formatting of phone numbers to the model.

# Why?

Previously it was possible to put in invalid numbers e.g.

```
4784529 <-- No area code
0208 <-- Just area code!
020804784529 <-- Spot the extra invalid digit?
     0208        4784529 <-- WT? Yes there were some like this.
442084784529 <-- an attempt at specifying an international dialling code
```

Now the number must be a valid UK telephone number including area code or a mobile number.

# The new rules

* Now the number must be valid according to the `uk_telephone_numbers` Gem which implements the rules stated here: http://www.area-codes.org.uk/formatting.php#programmers
* If the user has attempted to space their number in a way that is helpful for readability  then we prefer that (e.g. `0333 00 22 333` instead of `0333 002 2333`) otherwise we format using `uk_telephone_numbers`

# What about existing data?

* There are 9 records in the existing database that are not valid. We will manually clean them up. Until they are fixed they will simply show as-is on the directory.
* The remaining 3000+ are valid but a subset are not formatted exactly as we want just yet. We now have the choice to re-save all office records to correct that or let it happen naturally over time.

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -61,8 +61,9 @@ class Office < ActiveRecord::Base
     ]
   end
 
-  def telephone_number
-    super.try { |x| x.gsub(' ', '') }
+  def telephone_number=(new_phone_number)
+    return super if new_phone_number.nil?
+    super new_phone_number.gsub(/\s+/, ' ').strip
   end
 
   def full_street_address
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -60,8 +60,11 @@ class Office < ActiveRecord::Base
   end
 
   def telephone_number=(new_phone_number)
-    return super if new_phone_number.nil?
-    super new_phone_number.gsub(/\s+/, ' ').strip
+    super cleanup_telephone_number(new_phone_number)
+  end
+
+  def telephone_number
+    return format_telephone_number(cleanup_telephone_number(super))
   end
 
   def full_street_address
@@ -99,5 +102,13 @@ class Office < ActiveRecord::Base
     if telephone_number.nil? || !UKPhoneNumbers.valid?(telephone_number.gsub(' ', ''))
       errors.add(:telephone_number, I18n.t("#{model_name.i18n_key}.telephone_number.invalid_format"))
     end
+  end
+
+  def cleanup_telephone_number(telephone_number)
+    telephone_number.try { |t| t.gsub(/\s+/, ' ').strip }
+  end
+
+  def format_telephone_number(telephone_number)
+    telephone_number.try { |t| UKPhoneNumbers.format(t) || t }
   end
 end
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -1,4 +1,5 @@
 require 'uk_postcode'
+require 'uk_phone_numbers'
 
 class Office < ActiveRecord::Base
   include Geocodable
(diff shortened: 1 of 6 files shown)


munckymagik:
There's a related PR to add the new locale string to the self-service side of things is https://github.com/moneyadviceservice/rad/pull/343

app/models/office.rb
```diff
@@ -61,8 +59,12 @@ def field_order
     ]
   end
 
+  def telephone_number=(new_phone_number)
+    super cleanup_telephone_number(new_phone_number)
+  end
+
   def telephone_number
-    super.try { |x| x.gsub(' ', '') }
+    return format_telephone_number(cleanup_telephone_number(super))
```
urfolomeus: Am I right in saying that, if we do an up-front cleanup of existing telephone numbers, then the call to `cleanup_telephone_number` here is no longer required?

spec/models/office_spec.rb
```diff
@@ -67,6 +67,28 @@
     end
   end
 
+  describe '#telephone_number=' do
+    context 'when `nil`' do
```
urfolomeus: <nitpick>why backticks around nil?</nitpick> :)

spec/models/office_spec.rb
```diff
@@ -77,10 +99,19 @@
     end
 
     context 'when provided' do
-      before { office.telephone_number = ' 07715 930 457  ' }
+      it 'removes whitespace from the front and back' do
```
urfolomeus: Given that this is being done `#telephone_number=` I'm not sure that this is actually testing what you are intending it to test. If you set the telephone number without using the setter, then the getter's behaviour can be tested.

spec/models/office_spec.rb
```diff
@@ -114,21 +145,71 @@
...
-      context 'when badly formatted' do
-        before { office.telephone_number = 'not-valid' }
+          it { is_expected.to_not be_valid }
+        end
 
-        it { is_expected.to_not be_valid }
+        context 'when the input characters other than spaces or numbers' do
+          before { office.telephone_number = 'not-numbers-or-spaces' }
+
+          it { is_expected.to_not be_valid }
```
urfolomeus: Did we decide whether or not we were going to make these checks more strict to ensure that the errors were on telephone number and not on some other field?

app/models/office.rb
munckymagik: Yes as we discussed but I changed my mind. I felt it's cheap enough that we might as well put it in for now. This way we don't *have* to do the upfront clean-up and so there's no need to answer the questions

>  is it a rake task? Is it a migration?  Is it a one-off script? If so where do I put it?
> Can it just be a one-liner we run in the console? If so where do I record this so it doesn't get forgotten?

Right now all we have to do is do a standard deploy including re-indexing to ElasticSearch and job done.

urfolomeus: Fair enough :)

spec/models/office_spec.rb
munckymagik: Yes.

spec/models/office_spec.rb
munckymagik: Yes. `update_column` side-steps the setter.

urfolomeus: >.< sorry I read the deleted line as being included. Ignore me.

spec/models/office_spec.rb
munckymagik: The fixture is valid until we make our single change to a single field so I'm happy enough we're covered.

urfolomeus: That is currently true, yes, but future changes might change that. I would feel more comfortable if we were explicitly checking for the right error.

urfolomeus:
:+1:

cazrin:
:+1:

rnalexander:
:+1: :shipit:

PR #150, opened 2016-03-11
Bug fix: telephone number not showing for any remote firms

munckymagik:
The `FirmResult#telephone_number` method is obsolete. It should have been deleted as part of https://github.com/moneyadviceservice/mas-rad_core/pull/146. But we missed it because telephone numbers only come from offices now. But in the rad_consumer the remote firms still seems to get their telephone number from the FirmResult object.

Essentially this was causing phone numbers for remote firms not to appear as they are now pre-formatted in ElasticSearch and UKPhoneNumbers (Gem) doesn't cope with input numbers containing spaces and returns nil.

So instead we just expose the telephone number attribute directly.

### DIFF
--- a/lib/mas/firm_result.rb
+++ b/lib/mas/firm_result.rb
@@ -22,7 +22,8 @@ class FirmResult
     :adviser_qualification_ids,
     :ethical_investing_flag,
     :sharia_investing_flag,
-    :languages
+    :languages,
+    :telephone_number
   ]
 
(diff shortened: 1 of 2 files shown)


mconnell:
:+1: Good spot!

rnalexander:
:+1:

munckymagik:
Thanks guys

PR #151, opened 2016-03-17
Migrate session from cookie to db

JiggyPete:
We found when adding the 'Recently Visited Firms' functionality in rad_consumer that the data being added to the cookie could be too much.

We did try and trim down the amount of data being added to the cookie with this [PR](https://github.com/moneyadviceservice/rad_consumer/pull/288) however if the amount of data grew we'd have the same issue.

Migrating to having the session within the DB seemed like the more robust solution.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    mas-rad_core (0.0.106)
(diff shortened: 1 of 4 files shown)


mconnell:
:+1:

Tr4pSt3R:
:+1:

PR #154, opened 2016-04-29
[WIP] Add temp import_tables for FCA Import

JiggyPete:
This PR introduces three tables to facilitate the migration of the fca import script to being something that can be run from within the admin area of RAD. 

This will allow Sandrine to run the import rather than take up dev time.

### DIFF
--- a/app/models/lookup/import/adviser.rb
+++ b/app/models/lookup/import/adviser.rb
@@ -15,6 +15,12 @@ module Lookup
       def self.truncate_sql
         "TRUNCATE #{table_name};"
       end
+
+      def self.import_uploaded_fca_data
+        lookup_class = Lookup::Adviser
+        lookup_class.delete_all
+        ActiveRecord::Base.connection.execute("INSERT INTO #{lookup_class.table_name} (SELECT * FROM #{table_name});")
+      end
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/models/lookup/import/adviser.rb
```diff
@@ -0,0 +1,26 @@
...
+      def self.table_name
+        "lookup_import_#{super}"
+      end
+
+      def self.fca_import_copy_statement
+        "COPY #{table_name} (reference_number, name, created_at, updated_at) FROM stdin;"
+      end
+
+      def self.truncate_sql
+        "TRUNCATE #{table_name};"
```
munckymagik: We can access a Rails provided way of truncating tables:

https://github.com/rails/rails/blob/v4.2.5.2/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L261

```ruby
ActiveRecord::Base.connection.truncate(table_name)
```

JiggyPete: We can, however that will put the truncate in a different transaction to the copy.

oneKelvinSmith:
:thumbsup:

paulanthonywilson:
18 months old and superseded.

PR #155, opened 2016-12-16
[DEV APPROVED] Added migration for website column on office table

jriga:
[Card 7836](https://moneyadviceservice.tpondemand.com/entity/7836)

User Story: 
As a principal/adviser, I would like to have the option to add my own website under offices so that consumers can get more information about my firm and the type of services we provide.

Is required by [rad#401](https://github.com/moneyadviceservice/rad/pull/401)


slaywell:
👍

jongilbraith:
👍

PR #157, opened 2017-01-23
RAD Admin -Fix "advisors can't seem to register offices"

tomas-stefano:
# The problem

When the user add ambiguos address with the building name on it, the system returns the validation error message:

```
1. Address could not be found by the mapping service.
We could not find the coordinates of your office address to plot it on the map in the directory.
Please check and try again. Ensure that you have put the street address, including building number, in the ‘Street address’ field and, if you have a building name enter this into the  separate ‘Building name’ field.
If the details are correct and still cannot be found, please contact RADenquiries@moneyadviceservice.org.uk for assistance.
```

So the main problem that Google is not returning latitude and longitude for ambiguous address
that has building name on the address. This is a limitation of the Geocoding API listed here:

https://maps-apis.googleblog.com/2016/11/address-geocoding-in-google-maps-apis.html

# Possible solutions

So the possible solutions were:

* Change to the Places API like suggested on the link above;
* Change to Bing
* Add a field name called Building name
* Just send the postcode to validate the address on Google.

After conversation with the stakeholders, they asked to just send the postcode instead the address line 1 and address line 2.

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -67,8 +67,13 @@ class Office < ActiveRecord::Base
     return format_telephone_number(cleanup_telephone_number(super))
   end
 
+  # The Geocodable interface expect an object that responds to
+  # Geocodable#full_street_address. So, to respect the interface,
+  # we created internally the method postcode_only_address,
+  # to reflect what this object returns differently from the others.
+  #
   def full_street_address
-    [address_line_one, address_line_two, address_postcode, 'United Kingdom'].reject(&:blank?).join(', ')
+    postcode_only_address
   end
 
   def has_address_changes?
@@ -91,6 +96,10 @@ class Office < ActiveRecord::Base
   end
 
   private
+
+  def postcode_only_address
+    [address_postcode, 'United Kingdom'].join(', ')
+  end
 
   def postcode_is_valid
     if address_postcode.nil? || !UKPostcode.parse(address_postcode).full_valid?
--- a/spec/models/office_spec.rb
+++ b/spec/models/office_spec.rb
@@ -334,18 +334,18 @@ RSpec.describe Office do
   describe '#full_street_address' do
     subject { office.full_street_address }
 
-    it { is_expected.to eql "#{office.address_line_one}, #{office.address_line_two}, #{office.address_postcode}, United Kingdom"}
+    it { is_expected.to eql "#{office.address_postcode}, United Kingdom"}
 
     context 'when line two is nil' do
       before { office.address_line_two = nil }
 
-      it { is_expected.to eql "#{office.address_line_one}, #{office.address_postcode}, United Kingdom"}
+      it { is_expected.to eql "#{office.address_postcode}, United Kingdom"}
     end
 
     context 'when line two is an empty string' do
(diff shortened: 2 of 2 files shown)


norm:
I would alter the method name too, as it is no longer returns a full street address.

tomas-stefano:
I had the same opinion before but there are other objects (models) that respond to the same method protocol of the Geocodable/ModelGeocoder object.

tomas-stefano:
So what I wanted to say above that the other objects that responds to this method **possibly** can return the full address, so only this one that will not return anymore the street address. Wdyt?

Tr4pSt3R:
It is possible to tell Geocoder what method will return the geocoding object. 

Check out the documentation here: https://github.com/alexreisner/geocoder#activerecord

tomas-stefano:
@Tr4pSt3R No, it is not about that. Perhaps I should have explained better in order to everybody understands. The model Geocoder it is a module that expect an object that respond to full street address. You can see here: https://github.com/moneyadviceservice/mas-rad_core/blob/master/lib/mas/model_geocoder.rb#L3
Then we have not just the office model that respond to full street addres, we have another models as well, the other models **possibly**, **potentially** can return the full street address, **so what I am saying here, that I had the same opinion to change the method name, but when I see other objects that can return the full street address I decide to leave this name, since **possibly** only this object can return only the postcode. Is it makes sense to everyone?

norm:
Perhaps then create a new method `postcode_only_address` (or something similar) and have `full_street_address` call that, and add an inline comment as to why you have to keep the interface but it can't return a full street address.

Tr4pSt3R:
Here is what we have,
``` Geocoder.coordinates(geocodable.full_street_address)```

This is what it can become, 
``` Geocoder.coordinates(geocodable.postcode)```

when the method that returns the Geocodable object is renamed to an appropriate name such as `postcode`. `ModuleGeocoder` is only a wrapper that we have added. 

### Documentation
For geocoding, your model must provide a method that returns an address. This can be a single attribute, but it can also be a method that returns a string assembled from different attributes (eg: city, state, and country).

Next, your model must tell Geocoder which method returns your object's geocodable address:

```
geocoded_by :full_street_address   # can also be an IP address
after_validation :geocode          # auto-fetch coordinates
```

tomas-stefano:
Ok, I already understand this since the beginning, but what I am trying to explaining that the * geocodable* object can be a Firm, an Office or Adviser object (an both three respond to full street address and the other two **possibly can return the full street address**). An even the Office object, the one I changed can send the full street address in case we decide to move to use the Google Places API. Again I will say: I had the same opinion to change to postcode, but when I saw other objects using the module and see the purpose of the module I decide to leave the same method name.

Tr4pSt3R:
Thanks for clarifying that there are other classes such as Firm and Advisor which does not use a post-code only full street address. I will thumb it once you've implemented @norm's recommendation on the name(_postcode_only_address_) of the method that returns the geocodable object.

norm:
Not quite what I meant for a comment, but you can at least look up the commit to see the explanation. :)

👍

PR #159, opened 2017-01-25
[WIP] - Add compatibility on elastic search 2.x

tomas-stefano:
With elastic 2.0 is not allowed anymore to pass the field _id on the document.
So the first thing we notice when try to index firms was the error on
elastic search:

```
MapperParsingException[Field [_id] is a metadata field and cannot be added
inside a document. Use the index API request parameters.]
```

Since we can pass the id on the resource path when you're creating
the documents **one by one** and **not on bulk** then removing
from source and parsing from the root node makes elastic works in
the same way as before.

## Aditional improvement

In order to facilitate the development of this feature, It was added logs in debug mode every time the app makes a request to elastic search and leave there for the same reason.

### DIFF
--- a/app/serializers/firm_serializer.rb
+++ b/app/serializers/firm_serializer.rb
@@ -1,8 +1,7 @@
 class FirmSerializer < ActiveModel::Serializer
   self.root = false
 
-  attributes :_id,
-    :registered_name,
+  attributes :registered_name,
     :postcode_searchable,
     :telephone_number,
     :website_address,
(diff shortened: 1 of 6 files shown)


paulanthonywilson:
@tomas-stefano This is about a year old. Is it still relevant?

tomas-stefano:
Sorry, on the midst of notifications I lost this one. So this was an first step on trying to update elastic search on rad. I will close this PR.

paulanthonywilson:
👍 I think I'll take 30 minutes to get rid of some of these abandoned remote branches too.

PR #161, opened 2017-02-17
When sort node is nil returns 0 for closest adviser

tomas-stefano:
Elastic search don't return the sort node when findin by
firms/:id only by search all /firms/_search url.
Since both request are using the firm result, we need to
add a little flexibility on this particular node.

I included on the same commit the attribute writer for
closest adviser. This makes possible for some other place
I set the closest adviser (when a user passing the distance for example)

This is related to https://github.com/moneyadviceservice/rad_consumer/pull/316


Tr4pSt3R:
👍

jongilbraith:
👍

PR #163, opened 2017-05-25
adds CSV renderer for Adviser objs

Tr4pSt3R:
Supports this ticket: https://github.com/moneyadviceservice/rad/pull/406

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/mas-rad_core/163)

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -1,3 +1,5 @@
+require 'csv'
+
 class Adviser < ActiveRecord::Base
   include Geocodable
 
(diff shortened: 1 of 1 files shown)


willhall88:
any tests to support this code?

Tr4pSt3R:
@willhall88 there's a test here: https://github.com/moneyadviceservice/rad/blob/8070-advisors_list_export/spec/features/admin/export_advisers_list_spec.rb#L100

With that being said maybe I should add a unit test for this method. I shall do so.

tomas-stefano:
Second thought on that. Since this is only on the admin I would create a simple Ruby class on rad admin instead of moving to the model.

Just a fake example:

```ruby
class AdvisersListCsv
   def initialize(advisers)
      @advisers = advisers
   end

def self.to_csv(_ = {})
     CSV.generate do |csv|
        csv << ['Ref. Number', 'Name', 'Firm']
 
        @advisers.each do |adviser|
          csv << [adviser.reference_number, adviser.name, adviser.firm.registered_name]
        end
     end
   end
end
```

And then on controller you can pass this object to respond_with using the Responders feature.

I would close this issue and moving the code to the rad admin.

Tr4pSt3R:
Moving the implementation from here to the RAD repo.

PR #164, opened 2017-11-23
[DEV APPROVED] Upgrade rails version (4.2.10)

osharpe:
TP 8702 https://moneyadviceservice.tpondemand.com/entity/8702

This PR is upgrade the Rails version to 4.2.10


willhall88:
getting the same rspec errors as Travis. give me a shout if you want to pair to fix

osharpe:
Thanks Will. I wasn't getting those failures before, but it looks like the upgraded active_model_serializers (0.10.7) the root = false is not required.

willhall88:
cool. tests all passing now 👍 
Happy with this PR to be merged.
There are a lot of deprecation warnings in the tests, mainly down to factoryGirl, so another PR should be followed to address this. I'm happy to pick this up though once your work is done, if you like.

osharpe:
Ye we have this across quite a lot of Repo's. Would be good if we can follow up with new PR to resolve this. Thanks @willhall88

pmcostello:
👍

PR #168, opened 2018-01-15
[DEV APPROVED] Redis.connect is now replaced by Redis.new

paulanthonywilson:
The previous Gem upgrades had updated the Redis gem to 4.0.1, which no longer has the method `Redis.connect`. Now it is `Redis.new`.

This is loosely connected to https://moneyadviceservice.tpondemand.com/RestUI/Board.aspx#page=board/4771110243071511547&appConfig=eyJhY2lkIjoiREI4ODcwRjkxMDNDRTM2NTlBMzhDNTRBRTVBNUU1N0UifQ==&boardPopup=userstory/8699/silent but effectively means that the current version of `mas-rad_core` on Rubygems is unusable in production mode.

### DIFF
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -5,6 +5,6 @@ if Rails.env.production?
     lookup: :google,
     use_https: true,
     api_key: ENV['GOOGLE_GEOCODER_API_KEY'],
-    cache: Redis.connect(url: ENV['REDISTOGO_URL'])
+    cache: Redis.new(url: ENV['REDISTOGO_URL'])
   )
 end


tomas-stefano:
Can you provide more information about the impact, please? What "previous Gem upgrades"? Can you post a link to this work, please? And since the connect method does not exist ... is this a bug that is happening on production? And mostly important, were there no tests to the existing code base to catch this? Do you think is good to add one?

paulanthonywilson:
1. This is the previous work to upgrade the gems. https://github.com/moneyadviceservice/mas-rad_core/pull/164
1. It is not happening on production because, presumably, no-one has tried to deploy this latest version of the gem - the one with the security upgrades - to production.
1. I only speculate why the previous developer did not add tests back in 2015. I would guess that it's because this initialiser is only called in production mode, and tests run in test mode. Also that that team were under immense pressure.
1. I don't think it's a good idea to add one; the difficulties of adding a test for a production-only configuration outweigh the chances of this happening again. Production deploys will fail, without taking down the current production system, which is a test enough.

PR #172, opened 2018-01-24
[WIP] 8802: Fixes serialization and to_json of the Firm model

scott-ad-riley:
It's not clear which gem exactly caused the issue but this line now returns an object with the serialized firm inside of a top level "firm" key after calling `to_json`.

This was likely introduced as part of the rails upgrade here: #164 but (luckily) the Elasticsearch re-index has not been run.


scott-ad-riley:
Closing because the issue is not here, it's fixed by a PR in `rad_consumer`

PR #173, opened 2018-03-20
update gem dependancies

pmcostello:
to fix security vulnerability in loofah gem


margOnline:
👍

osharpe:
I have regressed PG Gem from 1.0.0 to 0.21.0, as we were getting errors on Travis build. Looks like 1.0.0 is more suited to rails 5. Suspect we will see similar issues on other repo's when updating Gems.

tomas-stefano:
👍

